PR #1, opened 2016-06-11
v0.0.1 of boilerplate

brikis98:
This is v0.0.1 of our new project generator, which I’ve decided to call “boilerplate” instead of “gruntwork-generator”.

Functionality implemented so far:
1. Use a yaml file to define variables (and perhaps other configs in the future). JSON is too syntax heavy and doesn’t allow comments. HCL doesn’t support objects nested in lists. YAML seems like a well-known standard that keeps things clean & simple.
2. Reading the values for variables from command line options, var files, and by prompting the user.
3. Generating a project by copying all files from the source dir to the destination dir, passing text files through the Go template engine to allow variables to be filled in and other dynamic behavior.
4. Support for custom template helpers. One is implemented now, called `snippet`, which lets you embed snippets of other files in the current file. Very handy for documentation.
5. Tests.
6. Examples.
7. Docs.

Please see the README in the `_docs` folder for details. 

Remaining work:
1. ~~Automatically build the binaries and push them to GitHub for each release.~~
2. ~~Auto generate the root `README.md` file from the `_docs/README.md` file on release.~~
3. Add lots more examples.

### DIFF
--- a/_docs/README.md
+++ b/_docs/README.md
@@ -269,3 +269,7 @@ inspiring many of the ideas in Boilerplate and so you can try out other projects
    ```
 1. Consider open sourcing boilerplate. To do that, we may also want to open source the
    `gruntwork-module-circleci-helpers` it uses in `circle.yml`.
+1. Support a list of `ignore` files in `boilerplate.yml` or even a `.boilerplate-ignore` file for files that should be
+   skipped over while generating code.
+1. Consider supporting different types for variables. Currently, all variables are strings, but there may be value
+   in specifying a `type` in `boilerplate.yml`. Useful types: string, int, bool, float, list, map.


_docs/README.md
```diff
@@ -0,0 +1,275 @@
...
+Boilerplate is a tool for generating files and folders ("boilerplate") from a set of templates.
+
+Example use cases:
+
+1. Create the scaffolding for a new project (e.g. like [html5boilerplate](https://html5boilerplate.com/))
+1. Fill in boilerplate sections in your code files, such as including a legal disclaimer or license at the top of each
+   source file, or updating a version number in a file after each build.
+1. Embed code snippets from actual source files in documentation. Most READMEs have code copy/pasted into them and that
+   code often has syntax errors or goes out of date. Now you can keep those code examples in normal source files which
+   are built & tested, and embed parts of those files dynamically in your docs.
```
josh-padnick, line 12: Very cool use case.

_docs/README.md
```diff
@@ -0,0 +1,275 @@
...
+Note: the README for this project is generated using boilerplate! Check out the templates for it in the
+[_docs folder](/_docs) and the build job configuration in [circle.yml](/circle.yml).
+
+## Example: creating a new template
+
+Create a folder called `website-boilerplate` and put a file called `boilerplate.yml` in it:
+
+```json
+{{snippet "../examples/website/boilerplate.yml"}}
+```
```
josh-padnick, line 23: I like the syntax. Not sure I'd use "```json" though.

josh-padnick, line 23: Oh, interesting, I just realized that `{{snippet ...}}` will be replaced by boilerplate, so I guess you want "```yml" here.

josh-padnick, line 23: Looks like this is already fixed in `master`.

_docs/README.md
```diff
@@ -0,0 +1,275 @@
...
+You can also run Boilerplate non-interactively, which is great for automation:
+
+```
+boilerplate \
+  --template-folder /home/ubuntu/website-boilerplate \
+  --output-folder /home/ubuntu/website-output \
+  --non-interactive \
+  --var Title="Boilerplate Example" \
+  --var WelcomeText="Welcome!" \
+  --var ShowLogo="y"
```
josh-padnick, line 69: I really like the syntax. I remember we discussed this before (though I forget the conclusion), but can't `--non-interactive` be inferred based on whether any additional vars still need to be specified? Or is the idea that you can error out that var X is missing even though user was expecting this to be non-interactive?

_docs/README.md
```diff
@@ -0,0 +1,275 @@
...
+Copying /home/ubuntu/website-output/logo.png
+```
+
+Of course, Boilerplate can be used to generate any type of project, and not just HTML, so check out the
+[examples](/examples) folder for more examples and the [Working with Boilerplate](#working-with-boilerplate) section
+for full documentation.
+
+## Install
+
+Download the latest binary for your OS here: [boilerplate {{.Version}}](/releases/tag/{{.Version}}).
```
josh-padnick, line 81: At present, this syntax doesn't make sense to me but let's see what happens as I read further.

josh-padnick, line 81: Pretty cool this is auto-generated. Something we could easily do in CI.

_docs/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+   ```java
+   String str = "this is not part of the snippet";
+
+   // boilerplate-snippet: foo
+   String str2 = "this is part of the snippet";
+   return str2;
+   // boilerplate-snippet: foo
+   ```
+
```
josh-padnick, line 223: Love the `snippet` concept.

_docs/README.md
```diff
@@ -0,0 +1,275 @@
...
+  templating for file copy and file contents. Interop with Yeoman. Does not support non-interactive mode.
+* [hugo](https://gohugo.io/): Static website generator written in Go. Uses Go templates. Very focused on generating
+  websites, HTML, themes, etc, and doesn't support interactive prompts, so it's not a great fit for other types of
+  project generation.
+* [jekyll](https://jekyllrb.com/): Static website generator written in Ruby. Huge community. Very focused on generating
+  websites, HTML, themes, etc, and doesn't support interactive prompts, so it's not a great fit for other types of
+  project generation.
+* [play-doc](https://github.com/playframework/play-doc): Documentation generator used by the Play Framework that allows
+  code snippets to be loaded from external files. Great for ensuring the code snippets in your docs are from files that
+  are compiled and tested, but does not work as a general-purpose project generator.
```
josh-padnick, line 250: Great summary.

_docs/README.md
```diff
@@ -0,0 +1,275 @@
...
+     start: echo "Boilerplate is generating files to $2 from templates in $1"
+     file: echo "Boilerplate is processing file $1"
+     end: echo "Boilerplate finished generating files in $2 from templates in $1"
+   ```
+1. Consider open sourcing boilerplate. To do that, we may also want to open source the
+   `gruntwork-module-circleci-helpers` it uses in `circle.yml`.
+1. Support a list of `ignore` files in `boilerplate.yml` or even a `.boilerplate-ignore` file for files that should be
+   skipped over while generating code.
+1. Consider supporting different types for variables. Currently, all variables are strings, but there may be value
+   in specifying a `type` in `boilerplate.yml`. Useful types: string, int, bool, float, list, map.
```
josh-padnick, line 275: These docs are outstanding.

circle.yml
```diff
@@ -0,0 +1,28 @@
...
+
+deployment:
+  release:
+    tag: /v.*/
+    commands:
+      # If a new release is tagged in GitHub, build the binaries and upload them to GitHub.
+      - build-go-binaries --app-name boilerplate --dest-path bin --ld-flags "-X main.VERSION=$CIRCLE_TAG"
+      - upload-github-release-assets bin/*
+      # Generate the latest README it in the root folder and commit it to master
+      - bin/boilerplate_linux_amd64 --template-folder _docs --output-folder . --non-interactive --var Version=$CIRCLE_TAG
```
josh-padnick, line 27: Love it.

cli/boilerplate_cli.go
```diff
@@ -0,0 +1,147 @@
...
+	app.Author = "Gruntwork <www.gruntwork.io>"
+	app.Usage = "A tool for generating files and folders (\"boilerplate\") from a set of templates"
+	app.UsageText = "boilerplate [OPTIONS]"
+	app.Version = version
+	app.Action = runApp
+
+	app.Flags = []cli.Flag {
+		cli.StringFlag{
+			Name: config.OPT_TEMPLATE_FOLDER,
+			Usage: "Generate the project from the templates in `FOLDER`.",
```
josh-padnick, line 60: How does  ``FOLDER`` get rendered here?

config/config.go
```diff
@@ -0,0 +1,118 @@
...
+func BoilerPlateConfigPath(templateFolder string) string {
+	return path.Join(templateFolder, BOILERPLATE_CONFIG_FILE)
+}
+
+// Load the boilerplate.yml config contents for the folder specified in the given options
+func LoadBoilerPlateConfig(options *BoilerplateOptions) (*BoilerplateConfig, error) {
+	configPath := BoilerPlateConfigPath(options.TemplateFolder)
+
+	if util.PathExists(configPath) {
+		util.Logger.Printf("Loading boilerplate config from %s", configPath)
```
josh-padnick, line 67: Would you say that we've settled on a standard way to do logging in our golang apps at this point? If not, anything you would change in retrospect about the `util.Logger` pattern?

config/config.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+}
+
+// Parse the given configContents as a boilerplate.yml config file
+func ParseBoilerplateConfig(configContents []byte) (*BoilerplateConfig, error) {
+	boilerplateConfig := &BoilerplateConfig{}
+
+	if err := yaml.Unmarshal(configContents, boilerplateConfig); err != nil {
+		return nil, err
+	}
```
josh-padnick, line 86: I feel like Go does a really nice job of parsing JSON and YAML files. Do you think other langs like Scala are equally clean or do you see this as a unique strength of Go?

config/config.go
```diff
@@ -0,0 +1,118 @@
...
+	return boilerplateConfig, nil
+}
+
+// Validate that the config file has reasonable contents and return an error if there is a problem
+func (boilerplateConfig BoilerplateConfig) validate() error {
+	for _, variable := range boilerplateConfig.Variables {
+		if variable.Name == "" {
+			return errors.WithStackTrace(VariableMissingName)
+		}
+	}
```
josh-padnick, line 101: Do you need to check that the val is non-empty? Or I guess that's not necessarily an error?

config/config_test.go
```diff
@@ -0,0 +1,168 @@
+package config
```
josh-padnick, line 1: Just a question on process. Do you use TDD to generate these test files, or do you write the tests immediately after the file under test?

config/config_test.go
```diff
@@ -0,0 +1,168 @@
...
+func TestParseBoilerplateConfigInvalid(t *testing.T) {
+	t.Parallel()
+
+	_, err := ParseBoilerplateConfig([]byte("not-a-valid-yaml-file"))
+
+	assert.NotNil(t, err)
+
+	unwrapped := errors.Unwrap(err)
+	_, isYamlTypeError := unwrapped.(*yaml.TypeError)
+	assert.True(t, isYamlTypeError, "Expected a YAML type error for an invalid yaml file but got %s", reflect.TypeOf(unwrapped))
```
josh-padnick, line 30: Will this code execute if `err != nil`? Doesn't `assert.NotNil(t, err)` fail the test immediately if `err != nil`?

config/config_test.go
```diff
@@ -0,0 +1,168 @@
...
+	t.Parallel()
+
+	_, err := LoadBoilerPlateConfig(&BoilerplateOptions{TemplateFolder: "../test-fixtures/config-test/invalid-config"})
+
+	assert.NotNil(t, err)
+
+	unwrapped := errors.Unwrap(err)
+	_, isYamlTypeError := unwrapped.(*yaml.TypeError)
+	assert.True(t, isYamlTypeError, "Expected a YAML type error for an invalid yaml file but got %s", reflect.TypeOf(unwrapped))
+}
```
josh-padnick, line 168: Test coverage seems pretty solid here.

config/variables.go
```diff
@@ -0,0 +1,161 @@
...
+	"fmt"
+	"github.com/gruntwork-io/boilerplate/util"
+	"github.com/gruntwork-io/boilerplate/errors"
+	"strings"
+	"io/ioutil"
+	"gopkg.in/yaml.v2"
+)
+
+// Get a value for each of the variables specified in boilerplateConfig. The value can come from the user (if the
+// non-interactive option isn't set), the default value in the default value in the config, or a command line option.
```
josh-padnick, line 13: s/the default value in the default value in the config/the default value in the config/

config/variables.go
```diff
@@ -0,0 +1,161 @@
...
+			return variables, err
+		}
+		variables[variable.Name] = value
+	}
+
+	return variables, nil
+}
+
+// Get a value for the given variable. The value can come from the user (if the non-interactive option isn't set), the
+// default value in the default value in the config, or a command line option.
```
josh-padnick, line 29: same string replace comment as above.

config/variables.go
```diff
@@ -0,0 +1,161 @@
...
+func getVariable(variable Variable, options *BoilerplateOptions) (string, error) {
+	valueFromVars, valueSpecifiedInVars := getVariableFromVars(variable, options)
+
+	if valueSpecifiedInVars {
+		util.Logger.Printf("Using value specified via command line options for variable '%s': %s", variable.Name, valueFromVars)
+		return valueFromVars, nil
+	} else if options.NonInteractive && variable.Default != "" {
+		// TODO: how to disambiguate between a default not being specified and a default set to an empty string?
```
josh-padnick, line 37: We could have the convention for the `yml` file that if `default:` is specified with a blank value, to record an empty-value escape-sequence like `##EMPTY-VAL##`. But yes, what if someone actually uses that string, so maybe better to do nothing for now until this is an issue.

config/config_test.go
josh-padnick, line 30: Sorry, I'm dumb. You're testing that `err != nil`.

glide.yaml
```diff
@@ -0,0 +1,9 @@
+package: github.com/gruntwork-io/boilerplate
+owners:
+- name: Gruntwork
+  homepage: http://www.gruntwork.io
+import:
+- package: github.com/urfave/cli
+- package: github.com/stretchr/testify/assert
+- package: github.com/go-errors/errors
+- package: gopkg.in/yaml.v2
```
josh-padnick, line 9: Oh didn't realize YAML support required a third-party lib.

integration-tests/examples_test.go
```diff
@@ -0,0 +1,59 @@
...
+			err := app.Run(strings.Split(command, " "))
+			assert.Nil(t, err, "boilerplate exited with an error when trying to generate example %s: %s", templateFolder, err)
+			assertDirectoriesEqual(t, expectedOutputFolder, outputFolder)
+		}
+	}
+}
+
+// Diffing two directories to ensure they have the exact same files, contents, etc and showing exactly what's different
+// takes a lot of code. Why waste time on that when this functionality is already nicely implemented in the Unix/Linux
+// "diff" command? We shell out to that command at test time.
```
josh-padnick, line 50: good call.

templates/template_helpers.go
```diff
@@ -0,0 +1,151 @@
...
+// {{snippet "../../foo/bar"}}
+//
+// However, this only works if boilerplate is called from the same folder as the template itself. To work around this
+// issue, this function can be used to wrap boilerplate template helpers to make the path of the template itself
+// available as the first argument to the helper. The helper can use that path to relativize other paths, if necessary.
+func wrapWithTemplatePath(templatePath string, helper TemplateHelper) func(...string) (string, error) {
+	return func(args ... string) (string, error) {
+		return helper(templatePath, args...)
+	}
+}
```
josh-padnick, line 40: Elegant solution to the relative path problem.

templates/template_processor.go
```diff
@@ -0,0 +1,135 @@
...
+		} else {
+			return processFile(path, templateFolder, outputFolder, variables)
+		}
+	})
+}
+
+// Copy the given path, which is in the folder templateFolder, to the outputFolder, passing it through the Go template
+// engine with the given set of variables as the data if it's a text file.
+func processFile(path string, templateFolder string, outputFolder string, variables map[string]string) error {
+	isText, err := util.IsTextFile(path)
```
josh-padnick, line 39: Any reason why you opted to put `IsTextFile()` in `util` instead of this package?

templates/template_processor.go
```diff
@@ -0,0 +1,135 @@
...
+
+// Return true if this is a path that should not be copied
+func shouldSkipPath(path string, templateFolder string) bool {
+	return path == templateFolder || path == config.BoilerPlateConfigPath(templateFolder)
+}
+
+// Render the template at templatePath, with contents templateContents, using the Go template engine, passing in the
+// given variables as data.
+func renderTemplate(templatePath string, templateContents string, variables map[string]string) (string, error) {
+	tmpl, err := template.New(templatePath).Funcs(CreateTemplateHelpers(templatePath)).Parse(templateContents)
```
josh-padnick, line 124: Didn't realize golang had native support for `Funcs`. Cool!

josh-padnick, line 124: Why do you think go requires the use of `Parse()` before `Execute()`? Why wouldn't the lang designers just omit `Parse()` completely?

util/file_test.go
```diff
@@ -0,0 +1,45 @@
...
+		{"text-file.tf", true},
+	}
+
+	for _, testCase := range testCases {
+		actual, err := IsTextFile(fmt.Sprintf("../test-fixtures/util-test/is-text-file/%s", testCase.file))
+
+		assert.Nil(t, err)
+		assert.Equal(t, testCase.isText, actual, "Incorrect classification for %s", testCase.file)
+	}
+}
```
josh-padnick, line 37: I suppose this test will only test the `guessMimeTypeUsingFileCommand` since we run this on Darwin (Mac OS X) and and linux in CircleCI. Since you said the other doesn't work so well, I guess that means it won't work that well in Windows. Is the only consequence of that, that we'll run binary data through the template parser which will slow things down, or something additional to worry about?

josh-padnick:
First, my apologies for the delay in reviewing this excellent code. 

I must say, this is a really great tool, and the code quality and test coverage is excellent here. I had mostly questions and curiosities but no major challenges of any kind. So, officially LGTM!

_docs/README.md
brikis98, line 69: Yea, the idea is to show an error.

_docs/README.md
brikis98, line 81: Already in CI!

cli/boilerplate_cli.go
brikis98, line 60: This is supported by the CLI library we're using. See here for docs: https://github.com/urfave/cli#placeholder-values

config/config.go
brikis98, line 67: We should use a better logging library than the one built into Go. We should be able to log at different levels (INFO, DEBUG, etc), allow users to pick a level (via cmd line flags or configuration), configure where logs get outputted to, do log rotation (for services), and so on.

These libraries seem popular:

https://github.com/op/go-logging
https://github.com/Sirupsen/logrus

config/config.go
brikis98, line 86: Depends on your definition of clean. If you like reflection, then Go's approach is fine. JVM languages have access to the same thing with JSON using a library like Jackson: https://github.com/FasterXML/jackson

For more compile-time driven serialization, Play offers a JSON library (https://www.playframework.com/documentation/2.5.x/ScalaJsonCombinators) that can do some magic via macros. Not sure Go will ever have anything quite like that.

I'm fine with the reflection based approach, so both Go and Java seem equivalent here.

config/config.go
brikis98, line 101: Which val are you referring to?

config/config_test.go
brikis98, line 1: When I'm building some new project--that is, where I'm doing exploratory coding because I don't yet know exactly what the project does or what it'll look like--I usually do the initial testing in the exact way the user will use that project. For example, for a tool you run manually on the command line (e.g. boilerplate), I'll test it manually on the command line. For a tool you use manually via a web UI (e.g. Atomic), I'll test it manually via the web UI. For a tool you use via an API (e.g. terratest), I'll test via programmatic tests. Of these, only the very last one is what purists would consider true TDD. However, I strongly believe in starting from the user's perspective and working backwards, and in many cases, TDD is not an accurate representation of the user's perspective.

Once the project looks right from the user's perspective, then I go back and clean up the code and fill in any missing automated tests. After that, I typically use TDD for all bug fixes and minor changes. However, large new features that require exploratory coding will follow the same pattern as above.

brikis98, line 1: It's also worth mentioning that I try to build projects up incrementally, so "exploratory coding" should never be more than a few hours, before you go back, clean up, and add tests. I wouldn't recommend doing 6 months of "exploratory coding" and then going back and adding tests :)

config/variables.go
brikis98, line 13: Fixed: https://github.com/gruntwork-io/boilerplate/commit/57efbc40952db5a1e3d215cdb55a99472c5c172e

config/variables.go
brikis98, line 29: Fixed: https://github.com/gruntwork-io/boilerplate/commit/c3df5a5cdcae4fd1d465083c00386047f4cf4c47

templates/template_processor.go
brikis98, line 39: I put it in `util/file.go` as it seemed to be a relatively reusable/generic operation related to files, as opposed anything related to templates.

config/variables.go
brikis98, line 37: Roger.

templates/template_processor.go
brikis98, line 124: Most likely because parsing is slow. That doesn't matter if you only use the template once, but if you are using it over and over again, it's nice to do the parsing just once, and get back a reusable object that has an efficient way to render using the already-parsed content.

util/file_test.go
brikis98, line 37: Yea, we'll just end up reading in lots of unnecessary junk (e.g. images, videos, binaries) otherwise.

cli/boilerplate_cli.go
josh-padnick, line 60: Got it, thanks.

config/config.go
josh-padnick, line 67: A case for limiting log levels to just ERROR and INFO: http://dave.cheney.net/2015/11/05/lets-talk-about-logging

Curious on your opinion on this.

config/config.go
josh-padnick, line 86: Interesting. I think you're referring to https://www.playframework.com/documentation/2.5.x/ScalaJsonAutomated for the macros. 

What are the downsides of using reflection? What's the advantage of a macro over reflection? How can you have compile-time driven serialization if the data you receive comes at runtme? Also, isn't use of a macro, which as I understand will generate code at compile-time, more likely to have bugs or edge cases than a native language support?

Sorry for all the questions, but I feel like this kind of functionality is foundational to CRUD apps and helps me better understand where Go vs. Scala are well-suited. Thx for your input here.

config/config.go
josh-padnick, line 101: My mistake, I was reading this at the time as if it were validating the vars themselves, but it's just validating the config file. Later on, I discovered you wrote some `TODOs` anyway about how disambiguate between legitimately empty values and ones that should be non-empty.

config/config_test.go
josh-padnick, line 1: This is really interesting, and I've added this to my Evernote. I especially like the philosophy of working backward from the actual user experience.

When I started doing TDD, I noticed I wrote my code differently, to be more easily testable. But I guess if you have that in mind as you code in the first place, maybe you don't need TDD to get that benefit. It does put a bit more of an onus to discipline yourself to add test coverage after the fact, but I still like the focus on UX.

Thanks for sharing.

templates/template_processor.go
josh-padnick, line 39: Ah, missed the file breakdown when I first wrote this comment. Thanks for clarifying.

templates/template_processor.go
josh-padnick, line 124: Ah, got it. Thx for the explanation.

config/config.go
brikis98, line 67: Gah, goddamn VCs making people think that being "contrarian" is a good thing. 

The truth about logging is very simple: when you're debugging tricky production issues, you almost always wish you had either more info or less info. You want more info when errors are happening without leaving any obvious signs; you want less info when the errors are being drowned out by noise. The ability to control log levels is one way to do that.

It's immensely useful in all servers. Normally, you set it to a high level (INFO or higher) to keep noise low, but when requests start failing, or you have concurrency bugs, or something is slow, you crank it up to DEBUG or even TRACE and kick yourself in the ass for not having included more log output so you could see exactly what the code was doing.

It's immensely useful in CLI apps. Normally, you set it to a high level (INFO or higher) so it doesn't distract users with lots of useless noise. But when users report bugs, you ask them to set it to DEBUG or TRACE so you can figure out what the hell is going on, and kick yourself in the ass for not having included more log output so you could see exactly what the code was doing.

I have never, ever, ever wished I had included less logging or regretted using various log levels. I have very, very often wished I had included more logging, especially at the DEBUG or TRACE level.

josh-padnick, line 67: I really appreciate this perspective. The blog post argued from a "reasoning" perspective, whereas your response is based in real-world experience. Very insightful. Thx for sharing!

config/config.go
brikis98, line 86: > What's the advantage of a macro over reflection?

I'm not sure of all of the reasons, but my understanding is that people want to avoid reflection because:
1. It can be slow. I suspect this is one of those performance myths that isn't true anymore in the real world, but if you google "Java reflection slow", you'll get lots and lots of results.
2. It bypasses the type system. If you use reflection, you can do, more or less, whatever you want, and the compiler won't do much about it. That means the serialization/deserialization code could have lots of bugs and the compiler won't help you catch them. I haven't found this to be a problem in real-world serialization code, but perhaps others have, and type system purists avoid reflection whenever they can.
3. Reflection is magical. It's hard to reason about it, it's hard to debug, and it's harder to customize its behavior to do exactly what you want. Again, I'm not sure this is a big problem in real-world serialization code, but functional programming purists tend to avoid reflection whenever they can.

Since Play is all about performance, type safety, and functional programming, it makes sense that they avoid reflection, at least for their Scala API. Their original JSON API had you manually specify how to parse JSON. Later, they removed a lot of that repetitive boilerplate with macros.

To be honest, having used their alternative solution, both with and without macros, it feels like a case of purity over practicality, and I tend to side with the reflection folks on this one. 

> How can you have compile-time driven serialization if the data you receive comes at runtme?

Obviously, the parsing of the data from user requests can only happen at run time. However, with the Play approach, you define a) your data model and b) how to parse arbitrary data into that data model at compile time. This means the compiler can check a lot more of that logic and, in theory, that logic can execute much faster. With the reflection approach, you define a data model at compile time, but the logic of how you convert the runtime data into your data model happens at runtime (via reflection magic) and so the compiler doesn't offer much help, and you may pay a performance penalty. In practice, I haven't seen much of an impact from these differences, especially with thoroughly battle-tested reflection libraries like Jackson, so the compile-time advantages seem mostly theoretical to me.

> Also, isn't use of a macro, which as I understand will generate code at compile-time, more likely to have bugs or edge cases than a native language support?

There is no "native language support". Either you write code that does parsing logic at compile time using macros or at runtime using reflection. Neither is easy to write or debug and both require a fairly deep understanding of the language. Arguably, the macro stuff is the hardest of all to get right, but I suspect the difference in bugginess is not significant.

josh-padnick, line 86: This is a really great explanation. Thank you for writing this!

PR #6, opened 2016-08-26
Support a configurable missing key action

brikis98:
This PR adds a new flag to boilerplate called `missing-key-action`. This flag determines what boilerplate will do if a template references a variable that is not defined:

* `error` (default): exit with an error. Making this the default helps to catch a lot of bugs and typos in your templates.
* `invalid`: render the text `<no value>`. 
* `zero`: render the zero value of the variable. For now, this will always be an empty string, but if we support types for variables, it will be different for each type.

### DIFF
--- a/cli/boilerplate_cli.go
+++ b/cli/boilerplate_cli.go
@@ -77,7 +77,7 @@ func CreateBoilerplateCli(version string) *cli.App {
 		},
 		cli.StringFlag{
 			Name: config.OPT_MISSING_KEY_ACTION,
-			Usage: "What to do if a template looks up a variable that is not defined. Must be one of: 'invalid' (render the text '<no value>'), 'zero' (render the zero value for the variable), or 'error' (return an error and exit immediately). Default: 'error'.",
+			Usage: fmt.Sprintf("What `ACTION` to take if a template looks up a variable that is not defined. Must be one of: %s. Default: %s.", config.ALL_MISSING_KEY_ACTIONS, config.DEFAULT_MISSING_KEY_ACTION),
 		},
 	}
 
--- a/cli/boilerplate_cli.go
+++ b/cli/boilerplate_cli.go
@@ -75,6 +75,10 @@ func CreateBoilerplateCli(version string) *cli.App {
 			Name: config.OPT_VAR_FILE,
 			Usage: "Load variable values from the YAML file `FILE`. May be specified more than once.",
 		},
+		cli.StringFlag{
+			Name: config.OPT_MISSING_KEY_ACTION,
+			Usage: "What to do if a template looks up a variable that is not defined. Must be one of: 'invalid' (render the text '<no value>'), 'zero' (render the zero value for the variable), or 'error' (return an error and exit immediately). Default: 'error'.",
+		},
 	}
 
 	return app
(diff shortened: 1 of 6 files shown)


brikis98:
@josh-padnick This is ready for review

josh-padnick:
I haven't reviewed the code yet, but at first blush, `invalid` seems like the wrong word to return a placeholder. "Invalid" makes me think that if something isn't valid, an error should be thrown. Otherwise, we're just marking something explicitly as empty. So I vote to use the word `placeholder` instead.

cli/boilerplate_cli.go
```diff
@@ -113,10 +117,20 @@ func parseOptions(cliContext *cli.Context) (*config.BoilerplateOptions, error) {
 		return nil, err
 	}
 
+	missingKeyAction := config.DEFAULT_MISSING_KEY_ACTION
+	missingKeyActionName := cliContext.String(config.OPT_MISSING_KEY_ACTION)
```
josh-padnick: Should this be `missingKeyActionValue`? I feel like the name would be `--missing-key-action` or something similar.

config/config.go
```diff
@@ -16,15 +16,56 @@ const OPT_OUTPUT_FOLDER = "output-folder"
...
+		if missingKeyName == keyName {
+			return MissingKeyAction(i), nil
+		}
+	}
+	return MissingKeyAction(-1), errors.WithStackTrace(InvalidMissingKeyAction(keyName))
+}
+
+// The names of the missing keys. Go doesn't have enums, so we have to roll our own based on this example:
+// https://gist.github.com/skarllot/102a5e5ea73861ff5afe
+var missingKeyNames = []string{}
```
josh-padnick, line 51: Is there a reason why the following would be an insufficient way to handle this?

```
type MissingKeyAction int

const (
        Invalid MissingKeyAction = iota
        ZeroValue
        ExitWithError
)
```

josh-padnick, line 51: Oh, I'm guessing it's because you want to convert these to strings. Could you just define make `MissingKeyAction` of type `string` and initialize it to a sane string literal?

josh-padnick:
Left minor comments but overall LGTM! Feel free to respond as desired and merge.

brikis98:
> I haven't reviewed the code yet, but at first blush, invalid seems like the wrong word to return a placeholder. "Invalid" makes me think that if something isn't valid, an error should be thrown. Otherwise, we're just marking something explicitly as empty. So I vote to use the word placeholder instead.

It's based on the names of the settings supported by [Go's templating system](https://golang.org/pkg/text/template/#Template.Option):

```
"missingkey=default" or "missingkey=invalid"
	The default behavior: Do nothing and continue execution.
	If printed, the result of the index operation is the string
	"<no value>".
"missingkey=zero"
	The operation returns the zero value for the map type's element.
"missingkey=error"
	Execution stops immediately with an error.
```

cli/boilerplate_cli.go
brikis98: Renamed to `missingKeyActionValue`

config/config.go
brikis98: It's the need to a) convert strings to enums and b) enums to strings. An alias for `string` would probably work, but it's customary for enums to be backed by an ordinal, so I went with  the approach found here: https://gist.github.com/skarllot/102a5e5ea73861ff5afe

brikis98:
Merging now!

PR #13, opened 2016-09-16
Arithmetic

brikis98:
This PR adds some basic arithmetic and loop helpers to boilerplate. This will make it a bit easier to perform basic math, loops, etc. 

I’ve also updated the existing math helpers to take any type of value and try to convert it to an int, rather than explicitly requiring an int or float64 or a specific type. When working inside a template, you often have a mix of strings, ints, floats, etc and having to do specific type conversions is not usually an option.

### DIFF
--- a/examples/docs/README.md
+++ b/examples/docs/README.md
@@ -18,3 +18,20 @@ Here is how to use the `snippet` helper to embed files or parts of files from so
 ```html
 {{snippet "../website/index.html"}}
 ```
+
+## Arithmetic
+
+Here is how you can use the arithmetic helpers to create a numbered list:
+
+{{ with $index := "0" -}}
+{{plus $index 1}}. Item
+{{plus $index 2}}. Item
+{{plus $index 3}}. Item
+{{- end }}
+
+And here is another way to do it using the slice helper:
+
+{{ range $i := (slice 1 4 1) -}}
+{{$i}}. Item
+{{ end -}}
+
--- a/integration-tests/examples_test.go
+++ b/integration-tests/examples_test.go
@@ -52,7 +52,7 @@ func TestExamples(t *testing.T) {
 // takes a lot of code. Why waste time on that when this functionality is already nicely implemented in the Unix/Linux
 // "diff" command? We shell out to that command at test time.
 func assertDirectoriesEqual(t *testing.T, folderWithExpectedContents string, folderWithActualContents string) {
-	cmd := exec.Command("diff", "-u", folderWithExpectedContents, folderWithActualContents)
+	cmd := exec.Command("diff", "-r", "-u", folderWithExpectedContents, folderWithActualContents)
 
 	bytes, err := cmd.Output()
 	output := string(bytes)
--- a/templates/template_helpers.go
+++ b/templates/template_helpers.go
@@ -55,6 +55,11 @@ func CreateTemplateHelpers(templatePath string) template.FuncMap {
 		"snakeCase": snakeCase,
 		"camelCase": camelCase,
 		"camelCaseLower": camelCaseLower,
+		"plus": wrapArithmeticFunction(func(arg1 float64, arg2 float64) float64 { return arg1 + arg2 }),
+		"minus": wrapArithmeticFunction(func(arg1 float64, arg2 float64) float64 { return arg1 - arg2 }),
+		"times": wrapArithmeticFunction(func(arg1 float64, arg2 float64) float64 { return arg1 * arg2 }),
+		"divide": wrapArithmeticFunction(func(arg1 float64, arg2 float64) float64 { return arg1 / arg2 }),
+		"slice": slice,
 	}
 }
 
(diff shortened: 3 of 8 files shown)


brikis98:
@josh-padnick Merging. Feedback welcome :)

examples/docs/README.md
```diff
@@ -17,4 +17,21 @@ Here is how to use the `snippet` helper to embed files or parts of files from so
...
+{{plus $index 2}}. Item
+{{plus $index 3}}. Item
+{{- end }}
+
+And here is another way to do it using the slice helper:
+
+{{ range $i := (slice 1 4 1) -}}
+{{$i}}. Item
+{{ end -}}
+
```
josh-padnick, line 37: Think you forgot the ``` here.

josh-padnick: Syntax is intuitive.

integration-tests/examples_test.go
```diff
@@ -52,7 +52,7 @@ func TestExamples(t *testing.T) {
 // takes a lot of code. Why waste time on that when this functionality is already nicely implemented in the Unix/Linux
 // "diff" command? We shell out to that command at test time.
 func assertDirectoriesEqual(t *testing.T, folderWithExpectedContents string, folderWithActualContents string) {
-	cmd := exec.Command("diff", "-u", folderWithExpectedContents, folderWithActualContents)
+	cmd := exec.Command("diff", "-r", "-u", folderWithExpectedContents, folderWithActualContents)
```
josh-padnick, line 55: I'm guessing `-r` means recursive directory diff check? Surprised neither of us caught this before.

templates/template_helpers.go
```diff
@@ -55,6 +55,12 @@ func CreateTemplateHelpers(templatePath string) template.FuncMap {
 		"snakeCase": snakeCase,
 		"camelCase": camelCase,
 		"camelCaseLower": camelCaseLower,
+		"plus": wrapFloatFloatToFloatFunction(func(arg1 float64, arg2 float64) float64 { return arg1 + arg2 }),
```
josh-padnick, line 58: Is `wrapFloatFloatToFloatFunction` a typo? Looks like this should be `wrapFloatToFloatFunction`.

josh-padnick: Oh I see, you're using two floats and returning one.

templates/template_helpers.go
```diff
@@ -167,26 +173,109 @@ func extractSnippetName(line string) (string, bool) {
...
+	}
+}
+
+// Convert the given value to a float64. Does a proper conversion if the underlying type is a number. For all other
+// types, we first convert to a string, and then try to parse the result as a float64.
+func toFloat64(value interface{}) (float64, error) {
+	// Because Go is a shitty language, we have to call out each of the numeric types separately, even though the
+	// behavior for almost all of them is identical. If we tried to do a case statement with multiple clauses
+	// (separated by comma), then the variable v would be of type interface{} and we could not use float64(..) to
+	// convert it.
```
josh-padnick, line 240: I was getting worried you wouldn't complain about Go in this PR, but my fears are now allayed.

templates/template_processor_test.go
```diff
@@ -95,6 +95,12 @@ func TestRenderTemplate(t *testing.T) {
 		{"Snake case test: {{ .Foo | snakeCase }}", map[string]string{"Foo": "foo BAR baz!"}, config.ExitWithError, "", "Snake case test: foo_bar_baz"},
 		{"Camel case test: {{ .Foo | camelCase }}", map[string]string{"Foo": "foo BAR baz!"}, config.ExitWithError, "", "Camel case test: FooBARBaz"},
 		{"Camel case lower test: {{ .Foo | camelCaseLower }}", map[string]string{"Foo": "foo BAR baz!"}, config.ExitWithError, "", "Camel case lower test: fooBARBaz"},
+		{"Plus test: {{ plus .Foo .Bar }}", map[string]string{"Foo": "5", "Bar": "3"}, config.ExitWithError, "", "Plus test: 8"},
+		{"Minus test: {{ minus .Foo .Bar }}", map[string]string{"Foo": "5", "Bar": "3"}, config.ExitWithError, "", "Minus test: 2"},
+		{"Times test: {{ times .Foo .Bar }}", map[string]string{"Foo": "5", "Bar": "3"}, config.ExitWithError, "", "Times test: 15"},
+		{"Divide test: {{ divide .Foo .Bar | printf \"%1.5f\" }}", map[string]string{"Foo": "5", "Bar": "3"}, config.ExitWithError, "", "Divide test: 1.66667"},
```
josh-padnick, line 101: Does boilerplate gracefully handle divide by 0 error?

josh-padnick: Same question for invalid slices.

josh-padnick:
Minor comments only. Cool feature, and looks good!

examples/docs/README.md
brikis98: What ```?

integration-tests/examples_test.go
brikis98: Yes. It means the test was only checking the files in the root directory. I caught this by doing TDD, updating a test in a way that I expected to fail before making any code changes, and then discovering the test did not fail.

templates/template_helpers.go
brikis98: Did I mention Go was a shitty language today? :)

templates/template_helpers.go
brikis98: Oh, yea, I guess I did :-D

templates/template_processor_test.go
brikis98: No, I let Go's built-in error handling take care of that. It seems like both of those should be "fail immediately" style errors, so I'm not sure of the benefit of wrapping them just to fail immediately anyway.

examples/docs/README.md
josh-padnick: Meaning, this isn't formatted for syntax.

brikis98: Note that this file is `examples/docs/README.md`. It's an example project used for demonstration purposes and testing. It uses real Go template syntax and shouldn't be wrapped in ```.

josh-padnick: Ah, duh. My scoping mechanism was operating on the wrong level :) I'm good here.

PR #14, opened 2016-09-16
Improve the prompt for variable values

brikis98:
This PR improves the way the boilerplate prompts the user for variable values. I largely copied the style used by Terraform. See the screenshot below.

screen shot 2016-09-16 at 5 46 53 pm

### DIFF
--- a/config/variables.go
+++ b/config/variables.go
@@ -81,7 +81,16 @@ func getVariableFromVars(variable Variable, options *BoilerplateOptions) (string
 
 // Get the value for the given variable by prompting the user
 func getVariableFromUser(variable Variable, options *BoilerplateOptions) (string, error) {
-	value, err := util.PromptUserForInput(formatPrompt(variable, options))
+	util.BRIGHT_GREEN.Printf("\n%s\n", variable.Description())
+	if variable.Prompt != "" {
+		fmt.Printf("  %s\n", variable.Prompt)
+	}
+	if variable.Default != "" {
+		fmt.Printf("  (default: %s)\n", variable.Default)
+	}
+	fmt.Println()
+
+	value, err := util.PromptUserForInput("  Enter a value")
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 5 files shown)


brikis98:

config/variables.go
```diff
@@ -81,7 +81,16 @@ func getVariableFromVars(variable Variable, options *BoilerplateOptions) (string
 
 // Get the value for the given variable by prompting the user
 func getVariableFromUser(variable Variable, options *BoilerplateOptions) (string, error) {
-	value, err := util.PromptUserForInput(formatPrompt(variable, options))
+	util.BRIGHT_GREEN.Printf("\n%s\n", variable.Description())
```
josh-padnick, line 84: Bright green!

config/variables_test.go
```diff
@@ -8,54 +8,6 @@ import (
...
-
-	variable := Variable{Name: "foo", Prompt: "custom", Default: "bar"}
-	options := &BoilerplateOptions{}
-
-	actual := formatPrompt(variable, options)
-	expected := "custom (default: 'bar')"
-
-	assert.Equal(t, expected, actual)
-}
-
```
josh-padnick, line 58: You didn't want to test against multiline strings?

brikis98:
> Can you provide more color (ha!) around your decision to just use a longer prompt versus introducing a new property like details?

I'm a fan of the way Terraform prompts for variables. I think it's clear, handles long and short prompts well, and makes it obvious what the action is. I don't see much reason to reinvent the wheel, so I figure we can go with that look and feel until we find something definitively better.

config/variables_test.go
brikis98: Multiline strings in what? The prompt?

josh-padnick:
Sounds good.

config/variables_test.go
josh-padnick: Ah, my bad, they're not multiline strings. I guess what I was really getting at is: why remove these tests completely versus updating them?

brikis98: To be honest, these were not very useful tests. They are checking a part of the UI, and whenever you do that, you run a few risks:
1. UIs change often, so you spend a lot of time updating test code.
2. UI code, by its very nature, is designed for human consumption, and not computer consumption. This makes it hard to check the output. In particular, adding terminal colors means inserting all sorts of funky strings into the output, which makes the test code harder to write.
3. Most importantly, UI tests can only capture whether the right data is presented. They cannot typically capture if it "looks good". 

Obviously, this whole thing is a judgment call, but I felt that the tests that I deleted here were providing minimal value, were already expensive to maintain and were unlikely to catch any real bugs. When faced with the option of making the tests even more expensive by being able to handle colors, or refactoring the UI code significantly to support this use case--which may go away at the next re-design--or just removing them, I chose the latter.

josh-padnick: Your reasoning makes sense. In fact, I was listening to a [podcast on effective unit testing](http://www.se-radio.net/2016/05/se-radio-episode-256-jay-fields-on-working-effectively-with-unit-tests/) and the guest actively encouraged the pruning of unit tests that had outlived their usefulness. So, sounds like you made the right call!

PR #15, opened 2016-09-16
Add support for interpolation in path names

brikis98:
This PR updates `boilerplate` so that you can use Go template syntax and boilerplate variables in the names of files and folders. 

This is very handy in cases where the paths are determined dynamically. For example, if you were using `boilerplate` to generate a Java project, your template folder could contain the path `com/{{.PackageName}}/MyFactory.java`. If you run `boilerplate` against this template folder and enter
"gruntwork" as the `PackageName`, you'd end up with the file `com/gruntwork/MyFactory.java`.

### DIFF
--- a/_docs/README.md
+++ b/_docs/README.md
@@ -281,7 +281,10 @@ non-binary file through the [Go Template](https://golang.org/pkg/text/template) 
 data structure.
 
 For example, if you had a variable called `Title` in your `boilerplate.yml` file, then you could access that variable
-in any of your templates using the syntax `{{"{{"}}.Title{{"}}"}}`.
+in any of your templates using the syntax `{{"{{"}}.Title{{"}}"}}`. You can even use Go template syntax and boilerplate
+variables in the names of your files and folders. For example, you could create a file called
+`{{"{{"}}.Title{{"}}"}}.txt` and if the user ran `boilerplate` and "foo" as the `Title`, you'd end up with the file
+`foo.txt`.
 
 #### Template helpers
 
--- a/templates/template_processor.go
+++ b/templates/template_processor.go
@@ -127,7 +127,7 @@ func processTemplateFolder(options *config.BoilerplateOptions, variables map[str
 			util.Logger.Printf("Skipping %s", path)
 			return nil
 		} else if util.IsDir(path) {
-			return createOutputDir(path, options)
+			return createOutputDir(path, options, variables)
 		} else {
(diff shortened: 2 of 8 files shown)


brikis98:

_docs/README.md
```diff
@@ -281,7 +281,10 @@ non-binary file through the [Go Template](https://golang.org/pkg/text/template)
 data structure.
 
 For example, if you had a variable called `Title` in your `boilerplate.yml` file, then you could access that variable
-in any of your templates using the syntax `{{"{{"}}.Title{{"}}"}}`.
+in any of your templates using the syntax `{{"{{"}}.Title{{"}}"}}`. You can even use Go template syntax and boilerplate
+variables in the names of your files and folders. For example, you could create a file called
+`{{"{{"}}.Title{{"}}"}}.txt` and if the user ran `boilerplate` and "foo" as the `Title`, you'd end up with the file
+`foo.txt`.
```
josh-padnick, line 287: Wow, cool feature!

templates/template_processor.go
```diff
@@ -160,15 +160,21 @@ func createOutputDir(dir string, options *config.BoilerplateOptions) error {
...
+// path contains boilerplate syntax, use the given options and variables to render it to determine the final output
+// path.
+func outPath(file string, options *config.BoilerplateOptions, variables map[string]string) (string, error) {
+	templateFolderAbsPath, err := filepath.Abs(options.TemplateFolder)
 	if err != nil {
 		return "", errors.WithStackTrace(err)
 	}
 
-	fileAbsPath, err := filepath.Abs(file)
+	interpolatedFilePath, err := renderTemplate(file, file, variables, options.OnMissingKey)
```
josh-padnick, line 172: Notably, you're using the `renderTemplate()` function as is. Cool! So, since we're now rendering paths on the file system, I suppose we're still expecting UTF-8 on a *nix box and likely Windows boxes as well? Also, do we need to filter out a `/` somehow since filenames shouldn't be allowed to have `/`?

josh-padnick: Based on https://support.apple.com/en-us/HT202808, it looks like you may want to error out if the proposed filename or folder name contains `:`, `/` or `\`. Otherwise, maybe issue a warning if any of the other properties is violated?

brikis98: Won't the OS itself provide that error?

PR #16, opened 2016-09-16
Exit with an error if the boilerplate.yml file is missing

brikis98:
This PR updates `boilerplate` to exit with an error if there is no `boilerplate.ml` file in the template folder. I’ve also added a flag to make this behavior configurable in case you want to use boilerplate without variables.

### DIFF
--- a/_docs/README.md
+++ b/_docs/README.md
@@ -127,6 +127,9 @@ The `boilerplate` binary supports the following options:
 * `--missing-key-action ACTION` (optional): What to do if a template looks up a variable that is not defined. Must
   be one of: `invalid` (render the text "<no value>"), `zero` (render the zero value for the variable), or `error`
   (return an error and exit immediately). Default: `error`.
+* `--missing-config-action ACTION` (optional): What to do if a template folder does not have a `boilerplate.yml` file.
+  Must be one of: `exit` (return an error and exit immediately) or `ignore` (log a warning and process the template
+  folder without any variables). Default: `exit`.
 * `--help`: Show the help text and exit.
 * `--version`: Show the version and exit.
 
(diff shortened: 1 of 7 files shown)


brikis98:

_docs/README.md
```diff
@@ -153,7 +156,8 @@ boilerplate --template-folder ~/templates --output-folder ~/output --var-file va
 #### The boilerplate.yml file
 
 The `boilerplate.yml` file is used to configure `boilerplate`. The file is optional. If you don't specify it, you can
-still use Go templating in your templates, but no variables or dependencies will be available.
+still use Go templating in your templates so long as you specify the `--missing-config-action ignore` option, but no
+variables or dependencies will be available.
```
josh-padnick, line 160: Technically, couldn't you use boilerplate's plugins (not sure the right word) with literals as well?

config/config_test.go
```diff
@@ -332,7 +333,18 @@ func TestLoadBoilerPlateConfigFullConfig(t *testing.T) {
 func TestLoadBoilerPlateConfigNoConfig(t *testing.T) {
 	t.Parallel()
 
-	actual, err := LoadBoilerPlateConfig(&BoilerplateOptions{TemplateFolder: "../test-fixtures/config-test/no-config"})
+	templateFolder := "../test-fixtures/config-test/no-config"
+	_, err := LoadBoilerPlateConfig(&BoilerplateOptions{TemplateFolder: templateFolder})
```
josh-padnick, line 337: OCD: Should this be `LoadBoilerplateConfig` (Boilerplate = one word)

_docs/README.md
brikis98: Once we have plugins, sure, there may be other cases where a `boilerplate.yml` file is not necessary. If that turns out to be a common case, we can even change the default value to `ignore`. However, as it is now, I had a few cases where I had large `boilerplate.yml` files, with many dependencies, and a few of the dependencies had typos in their template folder paths. I didn't notice this because boilerplate just logged a warning.

config/config_test.go
brikis98: Good point. I had a few instances of this. All fixed here: https://github.com/gruntwork-io/boilerplate/commit/94781854c7205d23f41c736bd9c4d704ed112763

_docs/README.md
josh-padnick: Roger.

PR #20, opened 2016-09-20
Add types to boilerplate

brikis98:
This PR adds support for types to boilerplate. You can now declare variables as string, int, float, bool, list, map, or enum, both as a way to validate the values, and so that you can use the corresponding helpers in Go template syntax (e.g. such as using `range` with a list). 

This is a major refactor. Along the way, I renamed the `prompt` parameter to `description`, since the actual prompt, after the UI update, is now always “Enter a value”.


_docs/README.md
```diff
@@ -361,26 +373,3 @@ inspiring many of the ideas in Boilerplate and so you can try out other projects
 * [play-doc](https://github.com/playframework/play-doc): Documentation generator used by the Play Framework that allows
   code snippets to be loaded from external files. Great for ensuring the code snippets in your docs are from files that
   are compiled and tested, but does not work as a general-purpose project generator.
-
-## TODO
```
brikis98, line 365: These TODOs are now either done, or filed as issues in the boilerplate repo.

examples/java-project/com/{{.CompanyName | dasherize | downcase}}/example/Example.java
```diff
@@ -0,0 +1,17 @@
...
+public class Example {
+  public static final int EXAMPLE_INT_CONSTANT = {{.ExampleIntConstant}}
+  public static final Map<> EXAMPLE_MAP = ImmutableMap.of({{range $index, $key := .ExampleMap | keys}}{{if gt $index 0}}, {{end}}"{{$key}}", "{{index $.ExampleMap $key}}"{{end}});
+
+{{- if .IncludeEnum}}
+
+  public enum ExampleEnum {
+    {{range $index, $enumName := .EnumNames -}}
+      {{if gt $index 0}}, {{end}}{{$enumName}}
+    {{- end }}
```
brikis98, line 14: This Java file is an example that shows the benefits of having types, such as being able to do loops over maps and lists.

templates/template_helpers.go
```diff
@@ -400,6 +402,20 @@ func slice(start interface{}, end interface{}, increment interface{}) ([]int, er
 	return out, nil
 }
 
+// Return the keys in the given map. This method always returns the keys in sorted order to provide a stable iteration
+// order.
+func keys(m map[string]string) []string {
```
brikis98, line 407: I added an extra helper here to make using Go's template engine more practical with maps. To be honest, I'm not too happy with Go templating. Like the rest of the Go language, it is missing a huge number of very basic features. Perhaps, in the future, we should move to a different templating engine.

test-fixtures/examples-expected-output/docs/README.md
```diff
@@ -22,7 +22,7 @@ Here is how to use the `snippet` helper to embed files or parts of files from so
   </head>
   <body>
     <h1>{{.WelcomeText}}</h1>
-    {{if eq .ShowLogo "true"}}<img src="logo.png">{{end}}
+    {{if .ShowLogo}}<img src="logo.png">{{end}}
```
brikis98, line 25: Another benefit of types: if-statements work correctly now.

util/collections.go
```diff
@@ -22,4 +24,43 @@ func ListContains(needle string, haystack []string) bool {
...
+
+	for key, value := range genericMap {
+		stringMap[ToString(key)] = ToString(value)
+	}
+
+	return stringMap
+}
+
+// Convert a generic map to a map from string to interface
+func ToStringToGenericMap(genericMap map[interface{}]interface{}) map[string]interface{} {
```
brikis98, line 52: Did I mention Go is a shitty language? Writing boilerplate code like this (ba-dum-tish) is going to drive me mad.

variables/dependencies.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// Given a map of key:value pairs read from a Boilerplate YAML config file of the format:
+//
+// name: <NAME>
+// template-folder: <TEMPLATE_FOLDER>
+// output-folder: <OUTPUT_FOLDER>
+//
+// This method takes the data above and unmarshals it into a Dependency object
+func UnmarshalDependencyFromBoilerplateConfigYaml(fields map[string]interface{}) (*Dependency, error) {
```
brikis98, line 89: The vast majority of time spent in this PR was in this manual unmarshalling code. The lack of a good type system in Go made this more painful than it needed to be.

variables/variables.go
```diff
@@ -0,0 +1,292 @@
+package variables
+
+import (
+	"fmt"
+	"github.com/gruntwork-io/boilerplate/util"
+)
+
+// An interface for a variable defined in a boilerplate.yml config file
+type Variable interface {
```
brikis98, line 9: The rest of the time in this PR was spent on this thing. It used to be a struct, but that led to lots of bugs where I was forgetting to set certain values in the struct (mainly, the new `Type` field). Since there is no way to enforce a default value in a struct, I converted it to an interface. 

In a better programming language, this would've been a perfect use case for multiple implementations of this interface: one for string variables, one for int variables, one for enums, and so on. Each implementation would only use the types relevant for it and would never return null or any of that nonsense for fields that didn't apply (e.g. only Enum has an `Options` field). In other words, you use the type system to make illegal types unrepresentable.

In go, this is somewhat possible, but it took a MASSIVE amount of code due to the lack of support for generics, self types, and other useful features you get in better type systems. The result hideous and definitely reduced the maintainability of the project, so I switched back to a single interface and a single implementation. It's not as good as it should be, but hopefully good enough for now.

brikis98:
@josh-padnick I'm going to merge this one in so I can add some follow-up work on top of it. Feedback welcome!

_docs/README.md
```diff
@@ -40,9 +40,19 @@ where you want the generated code to go:
 ```
 boilerplate --template-folder /home/ubuntu/website-boilerplate --output-folder /home/ubuntu/website-output
 
-Enter the value for 'Title': Boilerplate Example
-Enter the welcome text for the website: Welcome!
-Should the website show the logo? (y/n) (default: "y"): y
+Title
+
```
josh-padnick, line 44: Extra newline based on the below examples?

_docs/README.md
```diff
@@ -40,9 +40,19 @@ where you want the generated code to go:
...
+WelcomeText
+  Enter the welcome text for the website
+
+  Enter a value [type: string]: Welcome!
+
+ShowLogo
+  Should the website show the logo?
+
+  Enter a [type: bool]: true
```
josh-padnick, line 55: I like this new UX. The `description` can be as long as you want, and the prompt is typed. Nice!

_docs/README.md
```diff
@@ -66,7 +76,7 @@ boilerplate \
   --non-interactive \
   --var Title="Boilerplate Example" \
   --var WelcomeText="Welcome!" \
-  --var ShowLogo="y"
+  --var ShowLogo="true"
```
josh-padnick, line 79: Quotes still required?

_docs/README.md
```diff
@@ -164,7 +176,11 @@ variables or dependencies will be available.
 ```yaml
 variables:
   - name: <NAME>
-    prompt: <PROMPT>
+    description: <DESCRIPTION>
+    type: <TYPE>
+    options:
+      - <CHOICE>
+      - <CHOICE>
     default: <DEFAULT>
```
josh-padnick, line 184: Maybe a case can still be made for both `prompt` and `description`. For example, up above, you have the following example:

```
ShowLogo
   Should the website show the logo?

   Enter a [type: bool]: true
```

But is "Should the website show the logo?" really a `description`? That seems more like a `prompt` to me. What if the rendering looked something like this?

```
ShowLogo
   Should the website show the logo?

   Showing the logo helps promote the website but also takes more bandwidth.

   Enter a [type: bool]: true
```

The top text would be the `prompt` and the middle text the `description`. Also, I believe you're using colors in the terminal output now? You could make the description grey to further de-emphasize it.

_docs/README.md
```diff
@@ -174,41 +190,26 @@ dependencies:
...
-
-dependencies:
-  - name: about
-    template-folder: ../about-us-page
-    output-folder: ../about-us-page
-    variables:
-      - name: Description
-        prompt: Enter a description for the about page
-        default: About Us
+{{snippet "../examples/dependencies/boilerplate.yml"}}
```
josh-padnick, line 201: Nice. Are you using this example more than once, though?

_docs/README.md
```diff
@@ -285,7 +294,8 @@ non-binary file through the [Go Template](https://golang.org/pkg/text/template)
 data structure.
 
 For example, if you had a variable called `Title` in your `boilerplate.yml` file, then you could access that variable
-in any of your templates using the syntax `{{"{{"}}.Title{{"}}"}}`.
+in any of your templates using the syntax `{{"{{"}}.Title{{"}}"}}`. You can also use Go template syntax to do
+if-statements, for loops, and use the provided [template helpers](#template-helpers).
```
josh-padnick, line 298: Didn't realize markdown syntax requires escaping `{{`?

_docs/README.md
josh-padnick: Nice!

config/config_test.go
```diff
@@ -98,7 +109,113 @@ func TestParseBoilerplateConfigOneVariableMissingName(t *testing.T) {
...
+    options: foo
+`
+
+func TestParseBoilerplateConfigOneVariableEnumWrongType(t *testing.T) {
+	t.Parallel()
+
+	_, err := ParseBoilerplateConfig([]byte(CONFIG_ONE_VARIABLE_ENUM_OPTIONS_WRONG_TYPE))
+
+	assert.NotNil(t, err)
+	assert.True(t, errors.IsError(err, variables.InvalidTypeForField{FieldName: "options", ExpectedType: "List", ActualType: reflect.TypeOf("string"), Context: "foo"}), "Expected a InvalidTypeForField error but got %s", reflect.TypeOf(errors.Unwrap(err)))
```
josh-padnick, line 178: InvalidTypeForField because the `foo` variable has only one option?

examples/java-project/com/{{.CompanyName | dasherize | downcase}}/example/Example.java
josh-padnick: Wow, that's pretty sweet! We can actually auto-generate compilable, custom code with boilerplate!

examples/website/index.html
```diff
@@ -4,6 +4,6 @@
   </head>
   <body>
     <h1>{{.WelcomeText}}</h1>
-    {{if eq .ShowLogo "true"}}<img src="logo.png">{{end}}
+    {{if .ShowLogo}}<img src="logo.png">{{end}}
```
josh-padnick, line 7: Nice.

templates/template_helpers.go
josh-padnick: The missing feature in this case was sorting on the map's key values? Can you elaborate on what other limitations of Go's template engine you've run into?

util/collections.go
```diff
@@ -22,4 +24,43 @@ func ListContains(needle string, haystack []string) bool {
...
\ No newline at end of file
+}
+
+// Convert a generic list to a list of strings
+func ToStringList(genericList []interface{}) []string {
+	stringList := []string{}
+
+	for _, value := range genericList {
+		stringList = append(stringList, ToString(value))
+	}
```
josh-padnick, line 35: Should this code be more paranoid about errors thrown on `ToString()`?

util/collections.go
josh-padnick: It does feel a little bit like CS homework, doesn't it. I award an A for this func!

util/collections.go
josh-padnick: Ah, never mind, you're just using `fmt.Sprintf("%v", value)`.

variables/dependencies.go
josh-padnick: Given that the YAML could unmarshal into fields of various types, how would this kind of issues fared under a better-typed language like Scala? I guess you're saying you could use generics instead of individually writing out each of the different funcs?

variables/variables.go
josh-padnick: You make a valid point how a project like this, which now has first-class support for types, seems to fight against the limited type system of Go. I suppose we could consider working with other languages like Scala as well, but then we have all the overhead of setting up our SBT project, figuring out how to do CLI apps in Scala, one more language a future employee has to learn, etc.

So given the options, I think "good enough" is indeed good enough.

templates/template_helpers.go
brikis98: It's a "logic less" templating language, so you don't have access to the full power of a programming language. With such languages, you either need EVERYTHING to be meticulously filled out in your data blob, which in practical terms never works, or a huge slew of helpers, which it does not include. I've already had to add over a dozen helpers for basic tasks like arithmetic, text transforms, and even support for basic looping. If we stick with Go templating, to make boilerplate flexible and powerful, we'll have to add dozens more. 

I used to be a believer in logic less templating, especially after seeing what PHP devs would do in their "templates" (e.g. database calls), but the truth is that views DO have logic and SHOULD have logic and taking away a developer's tools for dealing with that logic is way too extreme of a solution.

variables/dependencies.go
brikis98: I actually haven't worked with YAML and Scala, but I have worked with JSON and Scala, and since those are similar, it means I'd have a bunch of advantages:
- Go doesn't allow you to set default values in structs. Therefore, after unmarshalling in Go, I'd end up with a bunch of blank fields, and would have to manually add validation logic all over the place to check for that and fill in defaults. In Scala, you can easily specify a default value.
- With reflection-based JSON libraries (e.g. [Jackson](https://github.com/FasterXML/jackson)), you can to specify a custom deserializer for a single type ([example](http://stackoverflow.com/a/19167145/483528)) or a single field within a type. In the deserializer, you implement a single method that returns the deserialized value. Simple, intuitive. With Go, you you can specify a custom deserializer for an entire type, but not individual fields. And when you implement it, the [API is very unintuitive](https://github.com/go-yaml/yaml/blob/v2/yaml.go#L32) and requires you to mess with pointers.
- With more manual JSON libraries (e.g. [Play JSON](https://www.playframework.com/documentation/2.5.x/ScalaJsonHttp)), you have strong tools for expressing how to parse each field in the JSON into the exact type you want. With go-yaml's clunky API, you don't.
- With macro-based JSON libraries (e.g. [Play JSON, again](https://www.playframework.com/documentation/2.5.x/ScalaJsonAutomated)), you can have a lot of the hard work done by a macro, and only do the custom stuff manually.
- Finally, due to Go's lack of support for generics and convenient methods for converting types, doing this sort of unmarshalling work involves a lot of convert `map[interface{}]interface{}` to `map[string]string` style work by hand.

variables/variables.go
brikis98: Unfortunately, the options for cross-platform CLI apps are very limited. Interpreted/dynamic languages like Ruby and Python are wonderful to work with, but a nightmare to distribute, as slightly different versions of Ruby, Python, and all their dependent libraries wreak havoc on portability. Compiled languages like Java and Scala include way too much overhead for simple CLI apps (e.g. you have to download 300 jars to run "Hello, World"). 

That leaves bash and Go. Bash is not a language built for anything over 100 lines, so it's not really an option for complicated apps. So that leaves Go.

The only other option is [Rust](https://www.rust-lang.org/en-US/). I haven't used it, but have heard very good things, and I _think_ it can do very lightweight, cross-platform binaries. My only concern is that it's relatively new and it's community is much smaller than Go's, so finding good 3rd party libraries would be a pain. It's supposed to be a fairly elegant and modern language, and hopefully the future of systems programming (a C/C++ replacement), but I'm not sure that's enough to warrant a switch right now.

_docs/README.md
brikis98: There is always a newline above the "Enter a value" text.

_docs/README.md
brikis98: Not strictly necessary here since it's one word, but a good practice to follow in general.

_docs/README.md
brikis98: Perhaps. I'm not convinced an extra parameter and juggling extra data offers a lot of benefit in this case though.

_docs/README.md
brikis98: It's used in the test code

_docs/README.md
brikis98: Not markdown, but boilerplate. That's because this README is, itself, processed by boilerplate as part of the CI job (note that it's in the `_docs` version, but the final version ends up in the root folder).

config/config_test.go
brikis98: The `options` field must be a list. If you wanted a list of one value, you would do:

``` yaml
options:
  - foo
```

Or using the shorter syntax:

``` yaml
options: [yaml]
```

PR #22, opened 2016-10-31
Fix bug with passing lists and maps to dependencies

brikis98:
This PR fixes a bug where boilerplate wasn’t able to pass lists and maps properly to dependencies. I updated the automated test to catch this bug. Before the fix, the test would exit with the error “Value '[Foo Bar Baz]' is not a valid value for variable 'EnumNames' with type 'list’.” After this fix, it works correctly.

The cause is as follows. When boilerplate gets a variable from YAML, a list is represented as an `[]interface{}` and a map as `map[interface{}]interface{}`. boilerplate looks for these types and converts them to `[]string` and `map[string]string`. However, when we passed this converted value to a dependency, we were still only looking for the more generic versions `[]interface{}` and `map[interface{}]interface{}`.

### DIFF
--- a/examples/dependencies-recursive/boilerplate.yml
+++ b/examples/dependencies-recursive/boilerplate.yml
@@ -26,3 +26,27 @@ dependencies:
 
         - name: Title
           description: Enter the title for the dependencies example
+
+    - name: java-project
+      template-folder: ../java-project
+      output-folder: ./java-project
+      variables:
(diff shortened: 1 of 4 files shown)


brikis98:
@josh-padnick Since you're traveling now, I'm going to merge this. Feedback welcome.

josh-padnick:
7 days later, LGTM!

PR #30, opened 2017-01-28
Add two new template helpers to support relative paths between dependencies.

josh-padnick:
While working with a client who had an unusual folder structure, I could set the `output-property` for any boilerplate dependencies to output my boilerplate file in the desired location, but computing relative paths to `teraform_remote_state` and for terragrunt dependencies wasn't possible. So this PR implements two new functions:

* **dependencyOutputFolderName DEPENDENCY_NAME STRING_TO_DELETE**: Return the value of the `output-folder` property of a dependency named `DEPENDENCY_NAME` with the substring `STRING_TO_DELETE` removed from it. 
* **dependencyOutputFolderRelPath BASE_DEPENDENCY_NAME TARGET_DEPENDENCY_NAME**: Return the relative path between the value of the `output-folder` property of the `BASE_DEPENDENCY_NAME` and `TARGET_DEPENDENCY_NAME`.

### DIFF
--- a/templates/template_helpers.go
+++ b/templates/template_helpers.go
@@ -458,32 +458,36 @@ func relPath(basePath, targetPath string) (string, error) {
 }
 
 // Find the value of the given property of the given Dependency.
-func boilerplateConfigDeps(options *config.BoilerplateOptions) func(string, string) string {
-	return func(name string, property string) string {
+func boilerplateConfigDeps(options *config.BoilerplateOptions) func(string, string) (string, error) {
+	return func(name string, property string) (string, error) {
 		deps := options.Vars["BoilerplateConfigDeps"].(map[string]variables.Dependency)
 		dep := deps[name]
 
+		if dep.Name == "" {
+			return "", fmt.Errorf(`The dependency "%s" was not found.`, name)
+		}
+
 		r := reflect.ValueOf(dep)
 		f := reflect.Indirect(r).FieldByName(property)
-		return f.String()
+		return f.String(), nil
 	}
 }
 
 // Find the value of the given property of the given Variable.
-func boilerplateConfigVars(options *config.BoilerplateOptions) func(string, string) string {
-	return func(name string, property string) string {
+func boilerplateConfigVars(options *config.BoilerplateOptions) func(string, string) (string, error) {
+	return func(name string, property string) (string, error) {
 		vars := options.Vars["BoilerplateConfigVars"].(map[string]variables.Variable)
 		myVar := vars[name]
 
+		if myVar.Name() == "" {
+			return "", fmt.Errorf(`The variable "%s" was not found.`, name)
+		}
+
 		r := reflect.ValueOf(myVar)
 		f := reflect.Indirect(r).FieldByName(property)
-		return f.String()
+		return f.String(), nil
 	}
 }
-
-//func boilerplateConfigDeps(name, property string) string {
-//
-//}
 
 // Custom errors
 


README.md
```diff
@@ -458,6 +458,10 @@ including conditionals, loops, and functions. Boilerplate also includes several
   file from which you are calling the `shell` helper. For another way to execute commands, see [hooks](#hooks).
 * `templateFolder`: Return the value of the `--template-folder` command-line option. Useful for building relative paths.
 * `outputFolder`: Return the value of the `--output-folder` command-line option. Useful for building relative paths.
+* `dependencyOutputFolderName DEPENDENCY_NAME STRING_TO_DELETE`: Return the value of the `output-folder` property of a 
+   dependency named `DEPENDENCY_NAME` with the substring `STRING_TO_DELETE` removed from it. 
+* `dependencyOutputFolderRelPath BASE_DEPENDENCY_NAME TARGET_DEPENDENCY_NAME`: Return the relative path between the 
+   value of the `output-folder` property of the `BASE_DEPENDENCY_NAME` and `TARGET_DEPENDENCY_NAME`.
```
brikis98: Docs will need to be updated

templates/template_helpers.go
```diff
@@ -67,6 +67,10 @@ func CreateTemplateHelpers(templatePath string, options *config.BoilerplateOptio
 		"shell": wrapWithTemplatePath(templatePath, shell),
 		"templateFolder": func() string { return options.TemplateFolder },
 		"outputFolder": func() string { return options.OutputFolder },
+		"dependencyOutputFolderRelPath": dependencyOutputFolderRelPath(rootConfig),
```
brikis98: Is this one still necessary?

templates/template_helpers.go
```diff
@@ -430,6 +434,137 @@ func shell(templatePath string, args ... string) (string, error) {
...
+
+		return relPath, nil
+	}
+}
+
+// TODO: Add description
+func boilerplateConfig(rootConfig *config.BoilerplateConfig) func(string, ...string) (string, error) {
+	if rootConfig == nil {
+		return nil
+	}
```
brikis98: What case does this cover?

templates/template_helpers.go
```diff
@@ -430,6 +434,137 @@ func shell(templatePath string, args ... string) (string, error) {
...
+		return "", fmt.Errorf("The dependency '%s' was not found!\n", dependencyName)
+	}
+
+	return val, nil
+}
+
+// TODO: Add description
+// TODO: Come up with a better name since this is just a string replace, not an actual prefix removal
+func stripPrefix(str, prefix string) string {
+	return strings.Replace(str, prefix, "", 1)
```
brikis98: https://golang.org/pkg/strings/#TrimPrefix

templates/template_helpers.go
```diff
@@ -430,6 +434,137 @@ func shell(templatePath string, args ... string) (string, error) {
...
+// TODO: Add description
+func boilerplateConfig(rootConfig *config.BoilerplateConfig) func(string, ...string) (string, error) {
+	if rootConfig == nil {
+		return nil
+	}
+
+	return func(configProperty string, identifiers ...string) (string, error) {
+
+		switch configProperty {
+		case "dependencies":
```
brikis98: Hmm, maintaining these manually will be a pain. But I agree reflection is not safe either. I'll post a comment shortly with some thoughts.

templates/template_helpers.go
```diff
@@ -430,6 +434,137 @@ func shell(templatePath string, args ... string) (string, error) {
...
+			switch propertyName {
+			case "default":
+				if variable.Default() != nil {
+					val = variable.Default().(string)
+				} else {
+					val = ""
+				}
+			case "description":
+				val = variable.Description()
+			case "type":
```
brikis98: Go's templating allows you to call functions and access fields in structs, so I don't think this level of detail is necessary. The API can probably be:

```
{{ (boilerplate-config "variables" "foo") .Default }} 
```

brikis98: Actually, this might have to be written:

```
{{ (boilerplate-config "variables" "foo") | .Default }} 
```

templates/template_helpers.go
josh-padnick: It's since been removed.

templates/template_helpers.go
josh-padnick: Replaced these with the approach you suggested in 5502446.

templates/template_helpers.go
josh-padnick: Thanks, updated to use `trimPrefix`. Also added a `trimSuffix` function.

templates/template_helpers.go
```diff
@@ -430,6 +437,54 @@ func shell(templatePath string, args ... string) (string, error) {
...
+
+		r := reflect.ValueOf(myVar)
+		f := reflect.Indirect(r).FieldByName(property)
+		return f.String()
+	}
+}
+
+//func boilerplateConfigDeps(name, property string) string {
+//
+//}
```
brikis98: Remove

templates/template_helpers.go
```diff
@@ -430,6 +437,54 @@ func shell(templatePath string, args ... string) (string, error) {
...
+func boilerplateConfigVars(options *config.BoilerplateOptions) func(string, string) string {
+	return func(name string, property string) string {
+		vars := options.Vars["BoilerplateConfigVars"].(map[string]variables.Variable)
+		myVar := vars[name]
+
+		r := reflect.ValueOf(myVar)
+		f := reflect.Indirect(r).FieldByName(property)
+		return f.String()
+	}
+}
```
brikis98: I still don't think these reflection-based functions are necessary with proper usage if the `index` helper.

brikis98: As I posted earlier:

Try `(index $.BoilerplateConfigDeps $depName).OutputFolder` instead: https://play.golang.org/p/R2aVeKwsQc

josh-padnick: > Try `(index $.BoilerplateConfigDeps $depName).OutputFolder` instead.

Thanks for this reminder, as I'd forgotten about this (perhaps with subconscious intention!). So, first, this does work as expected! It's a slight ding in terms of readability IMO:

```
{{- range $depName, $stringToDelete := .AlbsToInclude }}
    "{{ relPath ($.This.CurrentDep.OutputFolder) ((index $.BoilerplateConfigDeps $depName).OutputFolder) }}",
  {{- end }}
```

vs.

```
{{- range $depName, $stringToDelete := .AlbsToInclude }}
    "{{ relPath ($.This.CurrentDep.OutputFolder) (boilerplateConfigDeps $depName "OutputFolder") }}",
  {{- end }}
```

I didn't like reflection because you lose a lot of compile-time protections and runtime becomes a lot less predictable, but given the limited nature of these values, do you see any other concerns around using reflection?

Given that I've already implemented it as such and I like the syntax I'm inclined to keep it unchanged, but I'm open to other perspectives here.

brikis98: I'm not convinced it's worse in terms of readability. In fact, `index` is a documented, known Golang template function with an obvious API (map, key), whereas `boilerplateConfigDeps` is something custom that I have to check the docs for every time I use it.

But far more importantly than the subjective notion of readability is that `index` will fail in predictable ways as it's just a look up in a map, and when it fails, I'll know I just need to tweak my boilerplate template to fix it. `boilerplateConfigDeps` will fail in crazy ways, most likely with a panic, when something in the reflection code fails, and to fix it, I'll have to go and edit the boilerplate code, put up a PR, release a new version, etc... If I can even figure out the cause.

I know you wrote code around it already and it feels like a waste to remove it and change the usages of it, but boilerplate is becoming more and more important for us, and it's worth trying to keep it as maintainable as possible.

templates/template_helpers.go
josh-padnick: Thanks, removed.

templates/template_helpers.go
josh-padnick: > It's worth trying to keep it as maintainable as possible.

I'm definitely on board with that; sorry if I was getting a little whiny about making changes. ;-)

> But far more importantly than the subjective notion of readability is that index will fail in predictable ways as it's just a look up in a map, and when it fails, I'll know I just need to tweak my boilerplate template to fix it. 

In general, I agree with this logic, but in practice, ironically, it seems to be exactly the opposite. If I pass a non-existent dependency to the `index` function, it returns an empty value without erroring, whereas I can write code in the reflection function to set for a nil value and error accordingly, so now I'm actually thinking that is the better option. Thoughts?

brikis98: Haha, sure, go for it

josh-padnick: Added in 0c979f5.

PR #31, opened 2017-03-08
Add support for specifying env vars in shell helper and hooks

brikis98:
This PR allows you to specify environment variables when using the shell helper and hooks. This will be useful in a number of ways, but the main use case right now is to be able to specify a custom `TERRAGRUNT_SOURCE` environment variable so that as we are generating a project, Terragrunt uses the local copy of infrastructure-modules rather than trying to download something from Git that has not yet been committed/published.

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,10 +228,14 @@ hooks:
     - command: <CMD>
       args:
         - <ARG>
+      env:
+        <KEY>: <VALUE>
   after:              
     - command: <CMD>
       args:
         - <ARG>
+      env:
+        <KEY>: <VALUE>
 ```
 
 Here's an example:
@@ -371,7 +375,8 @@ Note the following:
 
 * The `before` hook allows you to run scripts before Boilerplate has started rendering.
 * The `after` hook allows you to run scripts after Boilerplate has finished rendering.
-* Each hook consists of a `command` to execute (required) plus a list of `args` to pass to that command (optional).
+* Each hook consists of a `command` to execute (required), a list of `args` to pass to that command (optional), and
+  a map of environment variables in `env` to set for the command (optional).
   Example:
    
     ```yaml
@@ -380,8 +385,10 @@ Note the following:
         args:
           - Hello
           - World
+        env:
+          FOO: BAR
     ```
-* You can use Go templating syntax in both `command` and `args`. For example, you can pass Boilerplate variables to 
+* You can use Go templating syntax in `command`, `args`, and `env`. For example, you can pass Boilerplate variables to 
   your scripts as follows:
     
     ```yaml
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -453,9 +460,11 @@ including conditionals, loops, and functions. Boilerplate also includes several
...
 * `keys MAP`: Return a slice that contains all the keys in the given MAP. Use the built-in Go template helper `.index`
   to look up these keys in the map.
-* `shell CMD`: Execute the given shell command and render whatever that command prints to stdout. The working directory
-  for the command will be set to the directory of the template being rendered, so you can use paths relative to the
-  file from which you are calling the `shell` helper. For another way to execute commands, see [hooks](#hooks).
+* `shell CMD ARGS...`: Execute the given shell command, passing it the given args, and render whatever that command 
+  prints to stdout. The working directory for the command will be set to the directory of the template being rendered, 
+  so you can use paths relative to the file from which you are calling the `shell` helper. Any argument you pass of the
+  form `ENV:KEY=VALUE` will be set as an environment variable for the command rather than an argument. For another way 
+  to execute commands, see [hooks](#hooks).
```
josh-padnick, line 467: How do you specify that you want `VALUE` to come for your system env vars? Would it be just `ENV:KEY=$VALUE`? May be worth clarifying in docs.

templates/template_helpers.go
```diff
@@ -25,6 +25,8 @@ var WHITESPACE_REGEX = regexp.MustCompile("[[:space:]]+")
 
 var PUNCTUATION_OR_WHITESPACE_REGEX = regexp.MustCompile("([[:space:]]|[[:punct:]])+")
 
+var ENV_VAR_REGEX = regexp.MustCompile("^ENV:(.+?)=(.*)$")
```
josh-padnick, line 28: Is the way to interpet the expression `(.+?)` "One or more characters, or none at all?" If that's the cause what's the difference between `(.+?)` and `(.*)`?

templates/template_helpers.go
```diff
@@ -422,12 +424,33 @@ func keys(m map[string]string) []string {
...
 		return "", errors.WithStackTrace(NoArgsPassedToShellHelper)
 	}
 
-	return util.RunShellCommandAndGetOutput(filepath.Dir(templatePath), args[0], args[1:]...)
+	args, envVars := separateArgsAndEnvVars(rawArgs)
+	return util.RunShellCommandAndGetOutput(filepath.Dir(templatePath), envVars, args[0], args[1:]...)
+}
+
+// To pass env vars to the shell helper, we use the format ENV:KEY=VALUE. This method goes through the given list of
+// arguments and split it into two lists: the list of cmd-line args and the list of env vars.
```
josh-padnick, line 437: s/split/splits

README.md
brikis98: Ah, that's not supported at the moment. I'll add a helper for it.

templates/template_helpers.go
brikis98: Ah, I can see how that's confusing. In regex, a question mark after a `+` or `*` means "non-greedy". It's a very common and useful pattern.

Consider the string `ENV:foo=bar=baz`.

If my regex was `ENV:(.+)=(.*)`, then the `.+` will be "greedy" and consume as much as it can until the final equals sign. The captured groups would be `["foo=bar", "baz"]`.

If my regex was `ENV:(.+?)=(.*)`, then the `.+?` will be "non-greedy" and stop consuming as soon as it hits the first equals sign. The captured groups would be `["foo", "bar=baz"]`.

templates/template_helpers.go
brikis98: Fixed: https://github.com/gruntwork-io/boilerplate/commit/5bffacaafae0b5d0bc0ddb1dfd40860c56fd3163

templates/template_helpers.go
josh-padnick: Ah, thanks for the clarification!

PR #32, opened 2017-03-08
Add helper to read environment variables

brikis98:
In a previous PR, I added a helper to *set* environment variables when using the `shell` helper or `hooks`. This PR adds a helper to *read* environment variables. The nice thing is that these helpers are composable: that is, you can use the helper to read env vars both with the `shell` helper and with `hooks`.

### DIFF
--- a/_docs/README.md
+++ b/_docs/README.md
@@ -202,10 +202,14 @@ hooks:
     - command: <CMD>
       args:
         - <ARG>
+      env:
+        <KEY>: <VALUE>        
   after:              
     - command: <CMD>
       args:
         - <ARG>
+      env:
+        <KEY>: <VALUE>
 ```
 
 Here's an example:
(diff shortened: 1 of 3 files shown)


_docs/README.md
```diff
@@ -202,10 +202,14 @@ hooks:
     - command: <CMD>
       args:
         - <ARG>
+      env:
+        <KEY>: <VALUE>
```
brikis98, line 206: Note: in the previous PR, I made my docs changes to the root README file. I forgot that in this repo, the README in the root folder is actually generated from the `_docs` folder, so I've had to redo those doc changes in the proper place in this PR :)

_docs/README.md
```diff
@@ -396,12 +402,16 @@ including conditionals, loops, and functions. Boilerplate also includes several
...
+* `shell CMD ARGS...`: Execute the given shell command, passing it the given args, and render whatever that command 
+  prints to stdout. The working directory for the command will be set to the directory of the template being rendered, 
+  so you can use paths relative to the file from which you are calling the `shell` helper. Any argument you pass of the
+  form `ENV:KEY=VALUE` will be set as an environment variable for the command rather than an argument. For another way 
+  to execute commands, see [hooks](#hooks).
 * `templateFolder`: Return the value of the `--template-folder` command-line option. Useful for building relative paths.
 * `outputFolder`: Return the value of the `--output-folder` command-line option. Useful for building relative paths.
-
+* `env NAME DEFAULT`: Render the value of environment variable `NAME`. If that environment variable is empty or not 
+  defined, render `DEFAULT` instead.
```
brikis98, line 413: This is the only change to the docs that's relevant for this PR

josh-padnick: +1 on implementing a `getOrElse()` here.

PR #55, opened 2019-12-17
Introduce a new helper function 'include'

yorinasub17:
This introduces `include`, a new helper function inspired by the same function in `helm`, which can be used as an alternative to `snippet`. The function works almost exactly like `snippet`, except instead of copying the contents verbatim, it can be used to pass the contents through the templating engine prior to including. This promotes reusability by having a common set of template text where we use boilerplate.

The primary use case is to refactor the `provider` code in the CIS compliance reference architecture so that we can reference variables like `AwsRegion` into a common snippet that is injected.


README.md
```diff
@@ -506,6 +506,11 @@ Boilerplate also includes several custom helpers that you can access that enhanc
    return str2;
    // boilerplate-snippet: foo
    ```
+* `include <PATH> <VARIABLES>`: Returns the contents of the file at `PATH` after rendering it through the templating
+  engine with the provided variables, as a string. Use `.` to pass the current variables to the included template. E.g:
```
brikis98, line 510: Might want to call out explicitly how this differs from `snippet` to make it extra clear

yorinasub17: Good idea! https://github.com/gruntwork-io/boilerplate/pull/56

PR #59, opened 2020-08-27
Remote templates support

yorinasub17:
This adds support in boilerplate to invoke remote templates by providing it with `go-getter` compatible URLs instead of paths. In the process I renamed all the places in the config and CLI flags that was using `template-folder` with `template-url` to match what we are expecting.

This means that we can now invoke templates in git, either as dependencies or directly on the CLI.

__NOTE__: This is a backwards incompatible change, but now (when we are getting ready for usage-patterns v2) is probably a good time to do this.

### DIFF
--- a/integration-tests/examples_test.go
+++ b/integration-tests/examples_test.go
@@ -29,13 +29,17 @@ func TestExamples(t *testing.T) {
 
 	outputBasePath, err := ioutil.TempDir("", "boilerplate-test-output")
 	require.NoError(t, err)
-	defer os.Remove(outputBasePath)
+	defer os.RemoveAll(outputBasePath)
 
 	examples, err := ioutil.ReadDir(examplesBasePath)
 	require.NoError(t, err)
 
 	for _, example := range examples {
 		if !example.IsDir() {
+			continue
+		}
+		if strings.Contains(example.Name(), "shell") {
+			// This is captured in TestExamplesShell
 			continue
 		}
 
@@ -69,29 +73,40 @@ func TestExamplesAsRemoteTemplate(t *testing.T) {
 
 	outputBasePath, err := ioutil.TempDir("", "boilerplate-test-output")
 	require.NoError(t, err)
-	defer os.Remove(outputBasePath)
+	defer os.RemoveAll(outputBasePath)
 
 	examples, err := ioutil.ReadDir(examplesBasePath)
 	require.NoError(t, err)
 
-	for _, example := range examples {
-		if !example.IsDir() {
-			continue
-		}
+	// Insulate the following parallel tests in a group so that cleanup routines run after all tests are done.
+	t.Run("group", func(t *testing.T) {
+		for _, example := range examples {
+			// Capture range variable to avoid it changing on each iteration during the tests
+			example := example
 
-		if example.Name() == "variables" {
-			t.Logf("Skipping example %s because it is implicitly tested via dependencies.", example.Name())
-			continue
-		}
+			if !example.IsDir() {
+				continue
+			}
+			if strings.Contains(example.Name(), "shell") {
+				// This is captured in TestExamplesShell
+				continue
+			}
 
-		t.Run(path.Base(example.Name()), func(t *testing.T) {
-			templateFolder := fmt.Sprintf("git@github.com:gruntwork-io/boilerplate.git//examples/%s?ref=%s", example.Name(), branchName)
-			outputFolder := path.Join(outputBasePath, example.Name())
-			varFile := path.Join(examplesVarFilesBasePath, example.Name(), "vars.yml")
-			expectedOutputFolder := path.Join(examplesExpectedOutputBasePath, example.Name())
-			testExample(t, templateFolder, outputFolder, varFile, expectedOutputFolder, string(options.ExitWithError))
-		})
-	}
+			if example.Name() == "variables" {
+				t.Logf("Skipping example %s because it is implicitly tested via dependencies.", example.Name())
+				continue
+			}
+
+			t.Run(path.Base(example.Name()), func(t *testing.T) {
+				t.Parallel()
+				templateFolder := fmt.Sprintf("git@github.com:gruntwork-io/boilerplate.git//examples/%s?ref=%s", example.Name(), branchName)
+				outputFolder := path.Join(outputBasePath, example.Name())
+				varFile := path.Join(examplesVarFilesBasePath, example.Name(), "vars.yml")
+				expectedOutputFolder := path.Join(examplesExpectedOutputBasePath, example.Name())
+				testExample(t, templateFolder, outputFolder, varFile, expectedOutputFolder, string(options.ExitWithError))
+			})
+		}
+	})
 
 }
 
--- a/integration-tests/examples_unix_test.go
+++ b/integration-tests/examples_unix_test.go
@@ -1,0 +1,59 @@
+// +build aix darwin dragonfly freebsd js,wasm linux netbsd openbsd solaris
+
+// The following tests should only be run on unix machines
+
+package integration_tests
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	"testing"
+
+	"github.com/gruntwork-io/boilerplate/options"
+	"github.com/gruntwork-io/terratest/modules/git"
+	"github.com/stretchr/testify/require"
+)
+
+func TestExamplesShell(t *testing.T) {
+	t.Parallel()
+
+	branchName := git.GetCurrentBranchName(t)
+	examplesBasePath := "../examples"
+	examplesExpectedOutputBasePath := "../test-fixtures/examples-expected-output"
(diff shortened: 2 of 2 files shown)


examples/dependencies-remote/boilerplate.yml
```diff
@@ -0,0 +1,32 @@
...
+    description: Enter the welcome text used by the website dependency
+
+  - name: ShowLogo
+    description: Should the webiste show the logo (true or false)?
+    type: bool
+    default: true
+
+dependencies:
+  - name: docs
+    template-url: "git@github.com:gruntwork-io/boilerplate.git//examples/docs?ref=master"
```
brikis98, line 21: Oh, interesting, I hadn't even considered dependencies being remote too!

getter-helper/getter_helper.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	return client, nil
+}
+
+// DownloadTemplatesToTemporaryFolder uses the go-getter library to fetch the templates from the configured URL to a
+// temporary folder and returns the path to that folder. If there is a subdir in the template URL, return the combined
+// path as well.
+func DownloadTemplatesToTemporaryFolder(templateUrl string) (string, string, error) {
+	workingDir, err := ioutil.TempDir("", "boilerplate-cache*")
```
brikis98: So we download to the system temp dir? We may want to call that out in the docs. Note that Terragrunt used to use the temp folder too, but we moved away from it and to a local `.terragrunt-cache` because:

1. The tmp folder often led to long file paths that worked poorly on Windows.
1. Having the files locally made debugging easier than hunting around in a temp folder.
1. We could avoid re-downloading the files on subsequent re-runs.

Not sure how many of these matter for boilerplate... 

Will we want a `--terragrunt-source` equivalent for local testing?

Also, what a `*` at the end of `boilerplate-cache` there?

Also

getter-helper/getter_helper.go
```diff
@@ -0,0 +1,162 @@
...
+	outDir := filepath.Clean(filepath.Join(cloneDir, subDir))
+
+	ctx, cancel := context.WithCancel(context.Background())
+	client, err := NewGetterClient(ctx, mainPath, cloneDir)
+	if err != nil {
+		cancel()
+		return workingDir, outDir, err
+	}
+
+	// Start getter in the background so we can trap and forward interrupt signals
```
brikis98: Why do a goroutine and channels here at all? Seems to make the implementation way more complicated, but the benefits aren't obvious to me.

options/options.go
```diff
@@ -22,7 +22,11 @@ const OptDisableShell = "disable-shell"
 
 // The command-line options for the boilerplate app
 type BoilerplateOptions struct {
-	TemplateFolder  string
+	// go-getter supported URL where the template can be sourced.
+	TemplateUrl string
+	// Working directory where the go-getter defined template is downloaded.
+	TemplateFolder string
```
brikis98, line 28: NIT: Perhaps `DownloadFolder` would be a clearer name?

getter-helper/getter_helper.go
yorinasub17: It's so you can `ctrl+c` during the git download (e.g., maybe it is hanging because of network connectivity) and have the cleanup routine run to delete the folder.

examples/dependencies-remote/boilerplate.yml
yorinasub17: Yup the nice use case for this is that customers (if we ever open this) can create their ref arch flavor that mixes gruntwork and company architecture catalogs. E.g., they can have a top level boilerplate driver that uses their flavor of VPC, but our flavor of EKS.

getter-helper/getter_helper.go
yorinasub17: > Not sure how many of these matter for boilerplate...

I think 1 and 3 matter (I don't see how having access to the templates locally is useful in 2), but I think there is a better solution for (1). I'll look into this.

3 sort of matters, but honestly I don't envision someone invoking the same template many times in one sitting. The use case for `boilerplate` is to scaffold a new environment/region/account and possibly component, which seems like it is something that you do once a day at most. Vs terragrunt is something you might run multiple times (`validate` then `plan`, and you might `plan` many resources, and then you will have to run `apply`, and possibly retry because terraform is buggy).

> Will we want a --terragrunt-source equivalent for local testing?

For now, this is not necessary because you can only specify the template via CLI. I think we should implement this when we implement the boilerplate config based CLI, where you can give it a config file instead of CLI args.

> Also, what a * at the end of boilerplate-cache there?

The `*` is how you tell `ioutil.TempDir` where to inject the unique ID in the folder name.

> Also ...

Looks like the last comment is incomplete. Was there another concern here with the tmp folder?

options/options.go
yorinasub17: Given the number of places this has to be updated and I don't think it is that misleading, I am going to leave it as is.

getter-helper/getter_helper.go
yorinasub17: After thinking about this, this is probably rare so removed: b7750e8

getter-helper/getter_helper.go
yorinasub17: Attempt at solving (1) [windows long path issue]: ea516e0

Will try it out on my windows box.

yorinasub17: Bad news: I tried to run this on windows, and I failed much earlier - I couldn't git clone the repo 😞 

See https://github.com/gruntwork-io/boilerplate/issues/60

examples/dependencies-remote/boilerplate.yml
brikis98: Neat 👍

getter-helper/getter_helper.go
brikis98: > > Not sure how many of these matter for boilerplate...
> 
> I think 1 and 3 matter (I don't see how having access to the templates locally is useful in 2), but I think there is a better solution for (1). I'll look into this.
> 
> 3 sort of matters, but honestly I don't envision someone invoking the same template many times in one sitting. The use case for `boilerplate` is to scaffold a new environment/region/account and possibly component, which seems like it is something that you do once a day at most. Vs terragrunt is something you might run multiple times (`validate` then `plan`, and you might `plan` many resources, and then you will have to run `apply`, and possibly retry because terraform is buggy).

I was thinking more of the person testing the code locally. That is, the author of the templates, not the user.

> > Will we want a --terragrunt-source equivalent for local testing?
> 
> For now, this is not necessary because you can only specify the template via CLI. I think we should implement this when we implement the boilerplate config based CLI, where you can give it a config file instead of CLI args.

What about for nested templates within `dependencies`? Again, thinking of a template author testing their code locally.

> > Also, what a * at the end of boilerplate-cache there?
> 
> The `*` is how you tell `ioutil.TempDir` where to inject the unique ID in the folder name.

Ah, got it!

> > Also ...
> 
> Looks like the last comment is incomplete. Was there another concern here with the tmp folder?

Hahah, whoops. Not sure. Hopefully nothing important 😁

brikis98: > Bad news: I tried to run this on windows, and I failed much earlier - I couldn't git clone the repo 😞
> 
> See #60

Doh.

yorinasub17: Ok I fixed the windows issue by supporting urlencoded characters in the file templates. This allowed me to clone the repo. I also fixed some windows specific test failures that were easy as well. See 89ca2b5

Of course, tests that depend on `bash` fail (e.g., the shell tests) but I feel pretty good about the windows tests in general. Once this merges into `master`, I can start looking into setting up CI for windows environments.

---

> I was thinking more of the person testing the code locally. That is, the author of the templates, not the user.

Fair enough. I'll start working on this functionality, but as a separate PR that merges into this branch given that I expect their will be a fair amount of code to add and probably some nuances to discuss that it deserves its own PR scope.

integration-tests/examples_test.go
```diff
@@ -69,29 +73,40 @@ func TestExamplesAsRemoteTemplate(t *testing.T) {
...
 
 	examples, err := ioutil.ReadDir(examplesBasePath)
 	require.NoError(t, err)
 
-	for _, example := range examples {
-		if !example.IsDir() {
-			continue
-		}
+	// Insulate the following parallel tests in a group so that cleanup routines run after all tests are done.
+	t.Run("group", func(t *testing.T) {
```
brikis98, line 82: Should there be a `t.Parallel()` in here? If not, does this run sequentially and then all other boilerplate tests after, in parallel?

yorinasub17: This needs to be synchronous so that the `defer` runs after this grouped test. This is a trick for having cleanup routines with parallel subtests (See "Cleaning up after a group of parallel tests" in the [subtests knowledge blog entry](https://blog.golang.org/subtests)).

getter-helper/getter_helper.go
yorinasub17: Filed https://github.com/gruntwork-io/boilerplate/issues/61 so we don't forget

PR #65, opened 2020-09-17
Add includes feature

bwhaley:
This PR adds an `includes` feature, allowing you to include templates within a template. This will help templates to be more DRY in the aws-architecture-catalog.

Right now I'm looking for feedback on the feature to make sure we're in agreement about how it should work. It may need some tweaking as I start using it while developing the architecture catalog.

### DIFF
--- a/README.md
+++ b/README.md
@@ -518,9 +518,8 @@ Foo variable is {{ .Foo }}
 ```
 
 The code above shows two configurations: `first-template` and `second-template`. The templates for both configurations
-call the variable `Foo`, but only `second-template` defines `Foo` with a default value. Even though `first-template`
-has a dependency defined for `second-template`, its variables (and their defualt values) are not available in
-`first-template`.
+call the variable `Foo`, but only `second-template` defines `Foo` with a default value. `Foo` (and its default value)
+is not available in `first-template`, even though `second-template` is a dependency.
 
 Now consider a similar configuration that uses includes instead:
 
@@ -547,12 +546,12 @@ $ cat second-template/example.txt
 Foo variable is {{ .Foo }}
 ```
 
-Using includes, the `Foo` variable is available to both `first-template` and `second-template`.
+Using includes, the `Foo` variable is only defined once, but it is available to both `first-template` and
+`second-template`.
 
-With includes, templates can be included *before* or *after* the current configuration. Unlike the `template-url`
-field of a dependency, included files are paths to a file, not a directory. This allows you to include multiple
-files from the same directory. Included templates can be either URLs or relative paths to a file. Here's another
-example:
+Templates can be included *before* or *after* the current configuration. Unlike the `template-url` field of a
+dependency, included files are paths to a file, not a directory. This allows you to include multiple files from
+the same directory. Included templates can be either URLs or relative paths to a file. Here's another example:
 
     ```yaml
     variables:
@@ -606,8 +605,9 @@ dependencies:
     output-folder: ../another/arbitrary/output/folder
 ```
 
-In this example, `before-dependency` is processed first, followed by `example-dependency` and finally
-`after-dependency`. Hooks follow this same convention.
+In this example, `before-dependency` and `after-dependency` are both included in the `dependencies` configuration. 
+`before-dependency` is processed first, followed by `example-dependency` and finally `after-dependency`. Hooks follow 
+this same convention.
 
 Note that included files can themselves include files.
 
--- a/_docs/README.md
+++ b/_docs/README.md
@@ -460,9 +460,8 @@ Foo variable is {{"{{"}} .Foo {{"}}"}}
 ```
 
 The code above shows two configurations: `first-template` and `second-template`. The templates for both configurations
-call the variable `Foo`, but only `second-template` defines `Foo` with a default value. Even though `first-template`
-has a dependency defined for `second-template`, its variables (and their defualt values) are not available in
-`first-template`.
+call the variable `Foo`, but only `second-template` defines `Foo` with a default value. `Foo` (and its default value)
+is not available in `first-template`, even though `second-template` is a dependency.
 
 Now consider a similar configuration that uses includes instead:
 
@@ -489,12 +488,12 @@ $ cat second-template/example.txt
 Foo variable is {{"{{"}} .Foo {{"}}"}}
 ```
 
-Using includes, the `Foo` variable is available to both `first-template` and `second-template`.
+Using includes, the `Foo` variable is only defined once, but it is available to both `first-template` and
+`second-template`.
 
-With includes, templates can be included *before* or *after* the current configuration. Unlike the `template-url`
-field of a dependency, included files are paths to a file, not a directory. This allows you to include multiple
-files from the same directory. Included templates can be either URLs or relative paths to a file. Here's another
-example:
+Templates can be included *before* or *after* the current configuration. Unlike the `template-url` field of a
+dependency, included files are paths to a file, not a directory. This allows you to include multiple files from
+the same directory. Included templates can be either URLs or relative paths to a file. Here's another example:
 
     ```yaml
     variables:
@@ -548,8 +547,9 @@ dependencies:
     output-folder: ../another/arbitrary/output/folder
 ```
 
-In this example, `before-dependency` is processed first, followed by `example-dependency` and finally
-`after-dependency`. Hooks follow this same convention.
+In this example, `before-dependency` and `after-dependency` are both included in the `dependencies` configuration. 
+`before-dependency` is processed first, followed by `example-dependency` and finally `after-dependency`. Hooks follow 
+this same convention.
 
 Note that included files can themselves include files.
 


README.md
```diff
@@ -335,6 +344,10 @@ See the [Dependencies](#dependencies) section for more info.
 
 See the [Hooks](#hooks) section for more info.
 
+**Includes**: Boilerplate can include the variables, dependencies, and hooks from another boilerplate template. Because order matters (for things like variable default values and hook execution), templates can be included *before* or *after* the current configuration.
```
yorinasub17, line 347: Will the includes feature include things like the templates themselves? E.g., if you set `includes.before`, is that the same thing as setting `dependencies` because it will also render that template? Or is it purely importing the metadata?

That might be worth clarifying here.

Side note: I think this should be purely the metadata to avoid confusion with `dependencies`, and avoiding the challenges of setting vars and output paths.

README.md
```diff
@@ -475,6 +489,38 @@ Note the following:
   dependencies.
 * For an alternative way to execute commands, see the `shell` helper in [template helpers](#template-helpers).
 
+#### Includes
+
+Boilerplate can include the configuration from one template inside another. Templates can be included *before* or
+*after* the current configuration. Unlike the `template-url`, included files are paths to a file, as opposed to a
```
yorinasub17: NIT: should clarify that this isn't any file, but a boilerplate yaml file.

```suggestion
*after* the current configuration. Unlike the `template-url`, included files are paths to a boilerplate yaml file, as opposed to a
```

config/config.go
```diff
@@ -68,7 +84,25 @@ func LoadBoilerplateConfig(opts *options.BoilerplateOptions) (*BoilerplateConfig
...
 
-		return ParseBoilerplateConfig(bytes)
+		boilerplateConfig, err := ParseBoilerplateConfig(bytes)
+		if err != nil {
+			return nil, errors.WithStackTrace(err)
+		}
+
+		// Reverse the order of the BeforeIncludes so that the usage is more intuitive.
+		// If we didn't reverse, the last item of the "before" list would be the first item included, but intuitively,
+		// one would expec tit to be the last item
```
yorinasub17:
```suggestion
		// one would expect it to be the last item
```

yorinasub17: This was super confusing to me as to why it happens, and I didn't get it until I read the code. Perhaps this should include a sentence that saids "this happens because in the before case, the items are prepended to the list, not appended"?

config/config.go
```diff
@@ -90,13 +124,79 @@ func ParseBoilerplateConfig(configContents []byte) (*BoilerplateConfig, error) {
...
+	}
+
+	absPath, err := filepath.Abs(path.Join(templateFolder, include))
+	if err != nil {
+		return "", "", errors.WithStackTrace(err)
+	}
+	return "", absPath, nil
+}
+
+// mergeIncludes merges an included file with the original configuration
```
yorinasub17:
```suggestion
// mergeIncludes merges a list of included files with the original configuration
```

config/config.go
```diff
@@ -90,13 +124,79 @@ func ParseBoilerplateConfig(configContents []byte) (*BoilerplateConfig, error) {
...
+		}
+
+		opts := &options.BoilerplateOptions{
+			TemplateFolder: path.Dir(includeFilePath),
+			IncludeFile:    includeFilePath,
+		}
+		srcConfig, err := LoadBoilerplateConfig(opts)
+		if err != nil {
+			return err
+		}
```
yorinasub17, line 174: Because this recursively calls `LoadBoilerplateConfig`, would this end up in an infinite loop if there is a chain of includes? For example,

A -inc-> B -inc-> C -inc-> A

util/collections.go
```diff
@@ -63,3 +63,13 @@ func ToStringToGenericMap(genericMap map[interface{}]interface{}) map[string]int
...
+
+// ReverseStringSlice returns a slice of strings in reverse order without modifying the original slice
+func ReverseStringSlice(src []string) []string {
+	dst := make([]string, len(src))
+	copy(dst, src)
+	for i, j := 0, len(dst)-1; i < j; i, j = i+1, j-1 {
+		dst[i], dst[j] = dst[j], dst[i]
+	}
+	return dst
+}
```
yorinasub17, line 75: NIT: Should there be unit tests for this?

config/config_test.go
```diff
@@ -608,6 +608,33 @@ func TestParseBoilerplateConfigMultipleHooks(t *testing.T) {
...
+		Includes: variables.Includes{
+			BeforeIncludes: []string{
+				"includes/before1.yml",
+				"includes/before2.yml",
+			},
+			AfterIncludes: []string{
+				"includes/after1.yml",
+			},
+		},
+	}
```
yorinasub17, line 632: NIT: Should also test how defaults are handled, and dependencies (in particular, making sure they are included in the right order).

README.md
bwhaley: Including a template will only include its variables, hooks, and dependencies, but will not render any templates. Will clarify in the docs!

config/config.go
bwhaley: I did wonder if this would be confusing. Will clarify.

config/config.go
bwhaley: Yes. The code attempts to detect this:

```
if includeFilePath == opts.IncludeFile {
    err := fmt.Errorf("Included file %s includes itself!", includeFilePath)
    return err
}
```

But this only works when one file includes itself. It won't be detected for URLs or for the scenario you describe. I've updated the README to warn about this because it won't be straightforward to detect this condition in a call cases.

config/config.go
bwhaley: aadc663

config/config.go
bwhaley: aadc663

config/config.go
bwhaley: aadc663

util/collections.go
bwhaley: aadc663

README.md
bwhaley: aadc663

README.md
bwhaley: aadc663

config/config.go
yorinasub17: We should file an issue to add cycle detection here. We've had a few cases where this happened in terragrunt and it threw many users off because it just hangs.

I think we can do rudimentary cycle detection using either a global (or scoped, where we pass forward a state tracking object) lookup table of visited nodes.

README.md
brikis98: I think the "but will not render any templates" needs more explanation:

- We should call that out explicitly, rather than implying it.
- If it's not rendering anything, what's the point of `includes`? Is mainly to reuse vars? In other words, can you give a real-world example of _why_ this feature is useful?
- If it includes `dependencies`... Do those render templates? Because `dependencies` normally do...
- How does `includes` compare to `dependencies`? It's worth explaining this explicitly, and discussing when you'd use one vs the other.

bwhaley: Tried to clarify in another update. See the relevant section [here](https://github.com/gruntwork-io/boilerplate/blob/1b17fd936b3a4c8edabe5dce7a4fa0be5ef1acbd/README.md#includes). Does this help to explain the feature and to disambiguate it from `dependencies`?

brikis98: I'm sorry, I probably missed a discussion that makes this all a lot more obvious, and while the updated description definitely helps, it's still not fully clicking with me. 

- Out of curiosity, what was the motivating use case? The actual thing you saw in `usage-patterns`, or you were planning on doing in `usage-patterns v2`, that made you want this feature?
- When you use `include` with something that has a `dependency`, are you just pulling in the variables and hooks of that dependency? Do the variables end up namespaced as they do with dependencies today? Do the hooks execute?
- Is the real, primary driver to re-use variable definitions? If so, perhaps we could extend `dependencies` with that ability? E.g., A `dependencies` block could specify `include: ["variables"]` to only include the variable definitions from a dependency or `include: ["variables", "hooks", "templates"]` to include everything, as it does now.
  
Again, really not trying to be stickler here, just feel like I'm missing something!

yorinasub17: My 2 cents is that if we are merging variables, we need more control over the ordering and the `includes` feature around `before` and `after` is very nice for that.

Also, here is the original discussion that spawned this: https://gruntwork-io.slack.com/archives/CLYE0JYS0/p1600278352135700

From there, my understanding is that we will be limiting our usage in `aws-architecture-catalog` to variables for now.

bwhaley: >Out of curiosity, what was the motivating use case? The actual thing you saw in usage-patterns, or you were planning on doing in usage-patterns v2, that made you want this feature?

The use case was raised as I was thinking about/working on the aws-architecture-catalog. It might help to look at my (very WIP) [branch](https://github.com/gruntwork-io/aws-architecture-catalog/tree/iac-1600).

tl;dr: Some variables are needed by many templates and in most cases will have the same default value. Without `includes`, the variable and its default must be defined in each `boilerplate.yml`, and there are a lot of these variables. It wasn't looking very DRY. See the `patterns/infrastructure-live/_includes` directory in that branch.

> When you use include with something that has a dependency, are you just pulling in the variables and hooks of that dependency? Do the variables end up namespaced as they do with dependencies today? Do the hooks execute?

If you include something with a dependency, those dependencies are just added to the other dependencies defined in the template. Same goes for hooks. So if you define a `before` include with hooks, its hooks are prepended to the hooks defined in the root template, and `after` includes are appended to the hooks defined in the root template. Then they all execute in order.

> Is the real, primary driver to re-use variable definitions? If so, perhaps we could extend dependencies with that ability? E.g., A dependencies block could specify include: ["variables"] to only include the variable definitions from a dependency or include: ["variables", "hooks", "templates"] to include everything, as it does now.

Yes, the primary driver is variable definitions. But I actually like the syntax of `includes` for this purpose. The branch I linked to is a good example. Right now, in that `_includes` directory, there are 5 include files, each with one or more variables. If these were managed as dependencies, we'd have:

```
dependencies:
  - name: dependency-one
    template-url: ../_includes/dependency-one
    output-folder: . # This field is unused - if we added the `include` attribute as you described, maybe this becomes optional

  - name: dependency-two
     template-url: ../_includes/dependency-two
     output-folder: .
... 3 more time ...
```

and we'd also have a directory structure like:

```
_includes
├── dependency-five
│   └── boilerplate.yml
├── dependency-four
│   └── boilerplate.yml
├── dependency-one
│   └── boilerplate.yml
├── dependency-three
│   └── boilerplate.yml
└── dependency-two
    └── boilerplate.yml
```

With includes, it's simplified to:
```
includes:
  before:
    - ../_includes/dependency-one.yml
    - ../_includes/dependency-two.yml
    - ../_includes/dependency-three.yml
    - ../_includes/dependency-four.yml
    - ../_includes/dependency-five.yml
```

With all the files in one directory.


The inclusion of hooks and dependencies in addition to variables is more of a byproduct. It didn't make sense to me to not include those as well, as intuitively, you'd think that if you're including a configuration, the entire configuration would be included. At least, that's what I'd expect of if I saw this feature in some tool.

Does that help?

brikis98: Ohhh, OK, thank you for the explanation. I think I was missing two pieces:

1. The concrete use case of all the repeated variables. These were indeed very non-DRY.
1. The fact that you `include` a `.yml` file directly, and not a folder that happens to include a `boilerplate.yml`.

Might be worth mentioning that in the docs.

This makes sense to me and I think I'm good with it.

I will say, separately, that I have some intuitive sense that there's some other way to solve this variable re-use problem, but I can't yet quite put my finger on it. I'll noodle on it in the background, but in the meantime, thanks for the explanation, and carry on 👍

bwhaley: Thanks Jim. Apologies that my explanations weren't very clear. I will endeavor to improve the docs.

> I have some intuitive sense that there's some other way to solve this variable re-use problem

That may well be the case. This solution has the benefit of also allowing reuse of hooks and dependencies, should the need arise. It's also a fairly clean implementation since we can just read in the included files before the processing happens.

One downside is that interpolation doesn't work so you can't use variables in the included paths. 

Anyway, I'll carry on with the aws-architecture-catalog stuff using includes for now.

PR #66, opened 2020-10-09
Add partials feature

bwhaley:
This introduces a new _layouts_ feature to boilerplate. For background, this comes from [an idea @brikis98 proposed](https://github.com/gruntwork-io/aws-architecture-catalog/pull/3#discussion_r499799567) on the `aws-architecture-catalog`. With this feature, a Boilerplate template can use the [`define` and `template`](https://golang.org/pkg/text/template/#hdr-Nested_template_definitions) features of Go's `text/template` package to define reusable templates that can wrap other content. For full details, here's a direct link to the [updated `README.md`](https://github.com/gruntwork-io/boilerplate/blob/template-layouts/README.md).

Note that in the discussion linked above, the hope was to use a boilerplate function to dynamically look up a template. Unfortunately, unless I'm missing something (which is a definite possibility!), that won't work, at least not in any easy/sane way that I can see. Template functions are run at template execution time (e.g. when [`Execute()`](https://golang.org/pkg/text/template/#Template.Execute) is called), meaning all the templates in the tree must already be parsed. We can't dynamically register another template in mid-`Execute`. Instead, we have to provide a map of layouts in the `boilerplate.yml` configuration, and pass those along to be `Parse`d and added to the template parse tree before running `Execute`.

I've also added a template helper function called `isTemplateDefined` to allow checking if a template is defined calling it. Without this, an undefined template throws an error, and we'd need multiple templates with a bunch of overlapping code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -337,7 +337,7 @@ executing the current one. Each dependency may contain the following keys:
 
 See the [Dependencies](#dependencies) section for more info.
 
-**Partials**: Use *partials* to include reusable templates. Partials are a defined using a list of glob patterns.
+**Partials**: Use *partials* to include reusable templates. Partials are defined using a list of glob patterns.
 
 * Globs are matched using [the Go `filepath.Glob` function](https://golang.org/pkg/path/filepath/#Glob)
 * In the event of a template name collision (e.g. multiple templates are defined with the same name), the last one wins.


README.md
```diff
@@ -328,6 +337,13 @@ executing the current one. Each dependency may contain the following keys:
 
 See the [Dependencies](#dependencies) section for more info.
 
+**Partials**: Use *partials* to include reusable templates. Partials are a defined using a list of glob patterns.
```
yorinasub17:
```suggestion
**Partials**: Use *partials* to include reusable templates. Partials are defined using a list of glob patterns.
```

README.md
```diff
@@ -615,6 +633,133 @@ to using the new naming scheme, as they will be updated to use the sprig version
     - `trimPrefixBoilerplate` and `trimSuffixBoilerplate`: Another name for the boilerplate versions of `trimPrefix`
       and `trimSuffix`. Use this if you would like to keep old behavior.
 
+#### Partials
+
+Partials help to keep templates DRY. Using partials, you can define templates in external files, and then use those, 
+templates over and over again in other templates. Partials are common among templating engines, such as in [Hugo](https://gohugo.io/templates/partials/).
```
brikis98:
```suggestion
Partials help to keep templates DRY. Using partials, you can define templates in external files, and then use those templates over and over again in other templates. Partials are common among templating engines, such as in [Hugo](https://gohugo.io/templates/partials/).
```

README.md
```diff
@@ -615,6 +633,133 @@ to using the new naming scheme, as they will be updated to use the sprig version
...
+Let's look at an example:
```
brikis98: NIT: Might be better to start with a simpler example. E.g., A `<head>...</head>` section that has a common set of meta tags included on every page and how you can `{{ define }}` it in a partial and use it in your templates via `{{ template }}`. Once the basic pattern is established, then show the more complicated layout example.

README.md
```diff
@@ -615,6 +633,133 @@ to using the new naming scheme, as they will be updated to use the sprig version
...
+</html>
+{{ end }}
+```
+
+Now, in each of the pages on the site, we can reuse this layout. For example, from the site's root, we want the welcome 
+page. We create the `boilerplate.yml` first:
+
+```yaml
+partials:
+  - ../../partials/layout.html
```
brikis98:
```suggestion
  - ../../partials/*.html
```

README.md
```diff
@@ -615,6 +633,133 @@ to using the new naming scheme, as they will be updated to use the sprig version
...
+```
+
+When we run `boilerplate`, `basic_layout` template will be rendered with the contents of the `index.html`. Then we can 
+use the same layout for the about page, with its corresponding `boilerplate.yml`.
+
+Contents of `about/boilerplate.yml`: 
+
+```yaml
+partials:
+  - ../../partials/layout.html
```
brikis98:
```suggestion
  - ../../partials/*.html
```

README.md
```diff
@@ -615,6 +633,133 @@ to using the new naming scheme, as they will be updated to use the sprig version
...
+You can use the template definitions from any of the included partials throughout your templates.
+
+You can use Go templating syntax in partial paths. For example, you can define a convenenience variable for a relative path to 
+make the paths slightly easier to read:
+
+```
+variables:
+  - name: TemplatesRoot
+    description: A convenience variable identify the relative path to the root of the templates directory.
+    default: ../../../../
```
brikis98, line 804: Neat 👍

README.md
bwhaley: 7efb073

README.md
bwhaley: 7efb073

README.md
bwhaley: 7efb073

README.md
bwhaley: 7efb073

PR #67, opened 2020-10-29
Fix issue with unmarshaling yaml maps

bwhaley:
This works around [an issue in the `yaml` package](https://github.com/go-yaml/yaml/issues/139) that causes YAML maps to be unable to be marshalled to JSON.

What was happening was that unmarshaling a map variable like this:

```
SomeMap:
  SomeKey: SomeVal
```

would result in a variable of type `map[interface{}]interface{}` rather than `map[string]interface{}`. The former cannot be marshaled to JSON correctly since the JSON package (and the JSON spec) require that keys are strings. Hence you could not marshal the map to JSON in a boilerplate template like this:

```
{{ .SomeMap | toPrettyJson }}
```

where [`toPrettyJson` is a function in sprig](https://masterminds.github.io/sprig/defaults.html). It would instead just render an empty string.

- [x] Needs a unit test

### DIFF
--- a/variables/yaml_helpers.go
+++ b/variables/yaml_helpers.go
@@ -282,7 +282,12 @@ func ParseYamlString(str string) (interface{}, error) {
 		return nil, errors.WithStackTrace(err)
 	}
 
-	return Convert(parsedValue), nil
+	parsedValue, err = ConvertYAMLToStringMap(parsedValue)
+	if err != nil {
+		return nil, errors.WithStackTrace(err)
+	}
+
+	return parsedValue, nil
 }
 
 // Parse a list of YAML files that define variables into a map from variable name to variable value. Along the way,
@@ -320,7 +325,17 @@ func parseVariablesFromVarFileContents(varFileContents []byte) (map[string]inter
 		return vars, err
 	}
 
-	return Convert(vars).(map[string]interface{}), nil
+	converted, err := ConvertYAMLToStringMap(vars)
+	if err != nil {
+		return nil, errors.WithStackTrace(err)
+	}
+
+	vars, ok := converted.(map[string]interface{})
+	if !ok {
+		return nil, YAMLConversionErr{converted}
+	}
+
+	return vars, nil
 }
 
 // Parse variables passed in via command line options, either as a list of NAME=VALUE variable pairs in varsList, or a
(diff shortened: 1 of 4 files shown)


variables/yaml_mapstr.go
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) 2015-2016 Michael Persson
+// Copyright (c) 2012–2015 Elasticsearch <http://www.elastic.co>
+//
+// Originally distributed as part of "beats" repository (https://github.com/elastic/beats).
+// Modified specifically for "iodatafmt" package.
```
yorinasub17: Is this still accurate? Perhaps we should have a link directly to the file in the repo where this came from?

variables/yaml_mapstr.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	// Base packages.
+	"fmt"
+
+	// Third party packages.
+	"gopkg.in/yaml.v2"
+)
+
+// Unmarshal YAML to map[string]interface{} instead of map[interface{}]interface{}.
+func Unmarshal(in []byte, out interface{}) error {
```
yorinasub17: NIT: This depends on the license, but can we rename this to something more descriptive, like `UnmarshalYaml`?

variables/yaml_mapstr.go
bwhaley: Updated in a37cb41 to point at the specific comment since the file in the beats repo is no longer present.

variables/yaml_mapstr.go
bwhaley: I was actually able to omit this due to other issues. See update in a37cb41.

variables/yaml_helpers.go
```diff
@@ -344,6 +342,25 @@ func ParseVars(varsList []string, varFileList []string) (map[string]interface{},
...
+		for k, v := range x {
+			m2[k.(string)] = Convert(v)
+		}
+		return m2
+	case []interface{}:
+		for i, v := range x {
+			x[i] = Convert(v)
+		}
+	}
+	return i
```
yorinasub17: NIT: we generally avoid single char variables as it makes the routines harder to read. Especially if the type is `interface{}`, and you reuse `i` (it's the `interface{}` function param, but also the forloop index). 

Might I suggest converting the following vars?

```
i => yamlMapOrList
x => typedYamlMapOrList
m2 => outputMap
```

variables/yaml_helpers.go
```diff
@@ -344,6 +342,25 @@ func ParseVars(varsList []string, varFileList []string) (map[string]interface{},
...
 
+// Convert updates i of type map[interface{}]interface{} to a map[string]interface{} so that it may be properly
+// marshalled in to JSON.
+// See: https://github.com/go-yaml/yaml/issues/139
+func Convert(i interface{}) interface{} {
+	switch x := i.(type) {
+	case map[interface{}]interface{}:
+		m2 := map[string]interface{}{}
+		for k, v := range x {
+			m2[k.(string)] = Convert(v)
```
yorinasub17: NIT: Should this do a type check so it won't panic if `k` is not a string? (defensive coding)

Something like:

```
strK, ok := k.(string)
if !ok {
  return nil, KeyIsNotStringErr{k}
}
m2[strK] = Convert(v)
```

variables/yaml_helpers_test.go
```diff
@@ -76,3 +76,44 @@ func TestParseVariablesFromKeyValuePairs(t *testing.T) {
...
+		},
+		{
+			input: []interface{}{
+				map[interface{}]interface{}{
+					"key1": "value1",
+				},
+				"",
+			},
+			expectedType: []interface{}{},
+		},
```
yorinasub17, line 112: NIT: Recommend adding another test with nested map, as well as using an alternative type for the value:

```
{
  "key3": 42,
  "key1": {
    "key2": "value2"
  }
}
```

variables/yaml_helpers.go
```diff
@@ -315,14 +314,13 @@ func ParseVariablesFromVarFile(varFilePath string) (map[string]interface{}, erro
...
 
-	err := yaml.Unmarshal(varFileContents, &vars)
-	if err != nil {
-		return vars, errors.WithStackTrace(err)
+	if err := yaml.Unmarshal(varFileContents, &vars); err != nil {
+		return vars, err
 	}
 
-	return vars, nil
+	return Convert(vars).(map[string]interface{}), nil
```
yorinasub17: NIT: This should do a type check first to avoid panic.

bwhaley: 6e977ec

variables/yaml_helpers_test.go
bwhaley: 6e977ec

variables/yaml_helpers.go
bwhaley: 6e977ec

variables/yaml_helpers.go
bwhaley: 6e977ec

PR #68, opened 2020-11-03
Add toYaml helper function

bwhaley:
Create a new helper function `toYaml` that will encode an input variable as YAML. This will be helpful to separate some configuration data out of HCL and in to YAML files. For example, a map of IAM users could be written as YAML in a `users.yml` file and included in a `terragrunt.hcl` with `yamldecode(file(users.yml))`.

### DIFF
--- a/_docs/README.md
+++ b/_docs/README.md
@@ -103,8 +103,8 @@ You can find older versions on the [Releases Page](https://github.com/gruntwork-
    helpers for common tasks such as loading the contents of another file.
 1. **Variable types**: Boilerplate variables support types, so you have first-class support for strings, ints, bools,
    lists, maps, and enums.
-1. **Scripting**: Need more power than static templates and variables? Boilerplate includes several hooks that allow 
-   you to run arbitrary scripts.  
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -571,6 +571,7 @@ Boilerplate also includes several custom helpers that you can access that enhanc
   defined, render `DEFAULT` instead.
 * `templateIsDefined NAME`: Returns a boolean indicating if template called `NAME` is known. Use this to conditionally 
   include one boilerplate template with another. Most often useful along with [partials](#partials).
+* `toYaml`: Encodes an input variable as a YAML string. Similar to the `toJson` function in sprig.
```
yorinasub17, line 574: I think you need to add this to https://github.com/gruntwork-io/boilerplate/blob/master/_docs/README.md. Then, on release, CI should render this boilerplate with the latest version.

bwhaley: Ahh, yes, of course. Done in 5565472 and fixed a few issues that were causing the deploy build step to fail, resulting in an out-of-date README.

PR #71, opened 2020-11-14
skip_files

yorinasub17:
This introduces `skip_files` to boilerplate templates, which allow you to specify files that boilerplate should conditionally ignore. This is useful for boilerplate documentation files, as well as creating templates that conditionally render different snippets based on user input (e.g., data template files for the architecture catalog).

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -693,7 +693,7 @@ func TestParseBoilerplateConfigWithSkipFiles(t *testing.T) {
 		Dependencies: []variables.Dependency{},
 		Hooks:        variables.Hooks{},
 		SkipFiles: []variables.SkipFile{
-			{Path: "docs/README_ALWAYS_SKIP.md", If: "true"},
+			{Path: "docs/README_ALWAYS_SKIP.md", If: ""},
 			{Path: "docs/README_MAYBE_SKIP.md", If: "{{ .MaybeSkip }}"},
 		},
 	}


README.md
```diff
@@ -488,6 +502,52 @@ Note the following:
...
+#### Skip Files
+
+You can specify files that should be excluded from the rendered output using the `skip_files` section in
+`boilerplate.yml`. This is most useful when you have templates that need to conditionally exclude files from the
+rendered folder list.
+
+The `skip_files` section is a list of objects with the fields `path` and `if`, where `path` is required. `if` can be
+used to conditionally skip a file from the template folder, and it defaults to `true`. That is, when `if` is omitted,
+the file at the path is always excluded from the output. Note that `path` is always the relative path from the template
+root. Both `path` and `if` support Go templating syntax with access to boilerplate [variables](#variables) and [template
```
brikis98, line 514: Does `path` support wildcards / globs?

templates/template_processor.go
```diff
@@ -219,6 +219,9 @@ func cloneOptionsForDependency(dependency variables.Dependency, originalOpts *op
 	}
 
 	templateUrl, templateFolder, err := options.DetermineTemplateConfig(renderedTemplateUrl)
+	if err != nil {
```
brikis98, line 222: Doh.

templates/template_processor.go
```diff
@@ -295,13 +298,33 @@ func shouldSkipDependency(dependency variables.Dependency, opts *options.Boilerp
 	return rendered == "true", nil
 }
 
+// Return true if the if parameter of the given SkipFile evaluates to a "true" value
+func skipFileIfCondition(skipFile variables.SkipFile, opts *options.BoilerplateOptions, variables map[string]interface{}) (bool, error) {
+	if skipFile.If == "" {
+		return false, nil
+	}
```
brikis98: I thought if you didn't specify `if` it would default to `true`?

variables/skip_files.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	// unmarshalStringField only returns nil pointer if there is an error, so we can assume it is not nil here.
+	path := *pathPtr
+
+	skipIfPtr, err := unmarshalStringField(fields, "if", false, path)
+	if err != nil {
+		return nil, err
+	}
+	// If the `if` attribute is omitted, we default it to true.
```
brikis98: Ohh... You default it to the string "true" in the parsing logic... Hm, I find that slightly unexpected TBH. It's a NIT, but for some reason, the "principle of least" surprise tells me that the parser should return `if` as empty string if the user didn't specify it, and the logic checking for the value of `if` treats empty string as a signal to default to a boolean `true`.

yorinasub17: Ok. I don't feel strongly about this so updated to what you mentioned: 9405f4d

templates/template_processor.go
yorinasub17: Updated in 9405f4d to what you suggest down below.

README.md
yorinasub17: Implemented glob support 85658d4

PR #77, opened 2021-09-20
not_path for skip

yorinasub17:
This implements `not_path` for the `skip` directive, which is the inverse of `path`. That is, `not_path` can be used to express "skip everything but the files that match".

This is useful for https://github.com/gruntwork-io/terraform-aws-architecture-catalog/pull/439, where we want to express "skip all files but the envcommon config".

### DIFF
--- a/templates/template_processor.go
+++ b/templates/template_processor.go
@@ -510,11 +510,24 @@ func anyNotPathDefined(skipFileList []ProcessedSkipFile) bool {
 }
 
 // pathInAnySkipNotPath returns true if the given path matches any one of the not_path attributes in the skip file list.
+// Note that unlike pathInAnySkipPath, this also does a directory check, where the directory is considered in the
+// not_path list. This is because the `not_path` list is an include list as opposed to a skip list, so we must copy over
+// the directory to copy over the exact file.
 func pathInAnySkipNotPath(canonicalPath string, skipFileList []ProcessedSkipFile) bool {
 	for _, skipFile := range skipFileList {
+		if skipFile.RenderedSkipIf == false || len(skipFile.EvaluatedNotPaths) == 0 {
+			continue
+		}
+
 		inSkipNotPathList := util.ListContains(canonicalPath, skipFile.EvaluatedNotPaths)
-		if skipFile.RenderedSkipIf && inSkipNotPathList {
+		if inSkipNotPathList {
 			return true
+		}
+
+		for _, path := range skipFile.EvaluatedNotPaths {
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -528,11 +528,17 @@ You can specify files that should be excluded from the rendered output using the
...
 
-The `skip_files` section is a list of objects with the fields `path` and `if`, where `path` is required. `if` can be
-used to conditionally skip a file from the template folder, and it defaults to `true`. That is, when `if` is omitted,
-the file at the path is always excluded from the output. Note that `path` is always the relative path from the template
-root. Both `path` and `if` support Go templating syntax with access to boilerplate [variables](#variables) and [template
-helpers](#template-helpers).
+The `skip_files` section is a list of objects with the fields `path`, `not_path`, and `if`, where one of `path` or
+`not_path` is required. When `path` is set, all files that match the `path` attribute will be skipped, while when
+`not_path` is set, all files that DO NOT match the `not_path` attribute are skipped (in other words, only paths that
+match `not_path` is kept).
```
brikis98, line 534: NIT:

```suggestion
`not_path` is set, all files that DO NOT match the `not_path` attribute are skipped (in other words, only paths that
match `not_path` are kept).
```

templates/template_processor.go
```diff
@@ -468,14 +468,67 @@ func shouldSkipPath(path string, opts *options.BoilerplateOptions, processedSkip
...
-	// First check if the path is a part of the the skipFile list. If the path matches with any entry in the skip files
-	// list and the if condition evaluates to true, skip the file.
-	for _, skipFile := range processedSkipFiles {
-		if skipFile.RenderedSkipIf && util.ListContains(canonicalPath, skipFile.EvaluatedPaths) {
-			return true
-		}
+	// First check if the path is a part of the the skipFile list. To handle skipping:
+	// - If the path matches with any entry in the skip files list and the if condition evaluates to true, skip the file.
+	// - If the path matches does NOT match with any entry in the not_path list, and the if condition evaluates to true,
+	//   skip the file.
```
brikis98, line 474: NIT: is this supposed to say "If the path does NOT match with any entry..."?

README.md
yorinasub17: Fixed in https://github.com/gruntwork-io/boilerplate/pull/78

templates/template_processor.go
yorinasub17: Fixed in https://github.com/gruntwork-io/boilerplate/pull/78

PR #79, opened 2021-09-23
Fix bug in not_path logic

yorinasub17:
This fixes a bug in the `not_path` logic, where it was ignoring the `if` condition. See comment below for more details on the bug.

### DIFF
--- a/templates/template_processor.go
+++ b/templates/template_processor.go
@@ -480,7 +480,7 @@ func shouldSkipPath(path string, opts *options.BoilerplateOptions, processedSkip
 		return true
 	}
 	// not in any == not in all
-	if anyNotPathDefined(processedSkipFiles) && pathInAnySkipNotPath(canonicalPath, processedSkipFiles) == false {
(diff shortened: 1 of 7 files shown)


templates/template_processor.go
```diff
@@ -499,10 +499,11 @@ func pathInAnySkipPath(canonicalPath string, skipFileList []ProcessedSkipFile) b
...
 }
 
-// anyNotPathDefined returns true if any skip file has a NotPath attribute defined.
-func anyNotPathDefined(skipFileList []ProcessedSkipFile) bool {
+// anyNotPathEffective returns true if any skip file has a NotPath attribute defined and the rendered if condition is
+// true.
+func anyNotPathEffective(skipFileList []ProcessedSkipFile) bool {
 	for _, skipFile := range skipFileList {
-		if len(skipFile.EvaluatedNotPaths) > 0 {
+		if skipFile.RenderedSkipIf && len(skipFile.EvaluatedNotPaths) > 0 {
```
yorinasub17, line 506: This check is necessary, because `pathInAnySkipNotPath` will return `false` when `RenderedSkipIf` is `false` for all skip directives with `not_path`.

PR #81, opened 2021-10-12
Fix template rendering for skip files and engines

yorinasub17:
The `path` and `not_path` attributes for `skip_files` and `engines` was not being rendered through boilerplate templating, when we want it to support that.

As a bonus, this adds more logging on `glob` failures so it is easier to understand what is going on. Otherwise, you get an unhelpful error message that simply states `file does not exist`.

### DIFF
--- a/templates/skip_files_processor.go
+++ b/templates/skip_files_processor.go
@@ -30,7 +30,7 @@ type ProcessedSkipFile struct {
 func processSkipFiles(skipFiles []variables.SkipFile, opts *options.BoilerplateOptions, variables map[string]interface{}) ([]ProcessedSkipFile, error) {
 	output := []ProcessedSkipFile{}
 	for _, skipFile := range skipFiles {
-		matchedPaths, err := renderGlobPath(opts, skipFile.Path)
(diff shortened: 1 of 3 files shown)


templates/skip_files_processor.go
```diff
@@ -94,13 +94,21 @@ func debugLogForMatchedPaths(sourcePath string, paths []string, directiveName st
...
+	rendered, err := render.RenderTemplateRecursively(opts.TemplateFolder, path, variables, opts)
 	if err != nil {
+		return nil, err
+	}
+
+	globPath := filepath.Join(opts.TemplateFolder, rendered)
+	rawMatchedPaths, err := zglob.Glob(globPath)
+	if err != nil {
+		// TODO: logger-debug - switch to debug
+		util.Logger.Printf("ERROR: could not glob %s", globPath)
```
brikis98, line 111: Switch to debug?

yorinasub17: We don't have logging levels implemented in `boilerplate` yet. This is a tag to help us track which ones we want to convert to debug logs when we implement it.
I thought we had an issue already for this, but filed a new one https://github.com/gruntwork-io/boilerplate/issues/82

Issue #89, opened 2022-02-23
Boilerplate is rendering escaped {{}}


yorinasub17:
We want to output the literal string `"artifacts/{{workflow.name}}"` without any template processing, but can't do it with the standard way go templates suggests of `"artifacts/{{ "{{workflow.name}}" }}"`, leading to an error where it tries to interpret the inner `{{}}` string.

The workaround is to individually escape `{{`, but that is very ugly:

```
"artifacts/{{"{{"}} workflow.name {{"}}"}}"
```

Any way we can make the first syntax work?

PR #92, opened 2022-02-23
Fix rendering issue when variable value contains templates syntax

yorinasub17:
## Description

Fix https://github.com/gruntwork-io/boilerplate/issues/89

The old implementation of recursively rendering the variables breaks a valid use case where you want to render a config file that relies on Go template syntax (e.g., Github Actions), as `boilerplate` will always try to render the nested template that we don't want rendered. This old implementation existed because we wanted to support nested variable lookups in the variable values.

To address this, I updated the variable rendering logic to use an approach similar to what we do in `terragrunt` with `locals`:

- Instead of recursively evaluating the template until we can't anymore, we do rendering _trials_ where we do single pass rendering attempts until there is no more errors, or no new variables are rendered.

This new approach should handle both use cases where we have multiple, nested variable references and the case where we want to output Go template syntax in a boilerplate variable value.

## TODOs

- [x] Update the docs.

### DIFF
--- a/config/get_variables.go
+++ b/config/get_variables.go
@@ -137,7 +137,6 @@ func getVariable(variable variables.Variable, opts *options.BoilerplateOptions) 
 		util.Logger.Printf("Using default value for variable '%s': %v", variable.FullName(), variable.Default())
 		return variable.Default(), nil
 	} else if opts.NonInteractive {
-		// TODO: this function does not make sense: this should handle missing key
 		return nil, errors.WithStackTrace(MissingVariableWithNonInteractiveMode(variable.FullName()))
 	} else {
 		return getVariableFromUser(variable, opts)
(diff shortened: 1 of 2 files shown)


config/get_variables.go
```diff
@@ -131,6 +137,7 @@ func getVariable(variable variables.Variable, opts *options.BoilerplateOptions)
 		util.Logger.Printf("Using default value for variable '%s': %v", variable.FullName(), variable.Default())
 		return variable.Default(), nil
 	} else if opts.NonInteractive {
+		// TODO: this function does not make sense: this should handle missing key
```
yorinasub17: In code review session with @zackproser we identified a potential bug where this change might no longer be handling the `--on-missing-key` option correctly. This needs a regression test and investigation to confirm.

yorinasub17: I looked into this further, and realized that `--on-missing-key` is for handling when the template references a variable that is not explicitly defined, and does not have a value for it provided via `--var` or `--var-file`, not when an explicitly defined variable is missing a value.

As such, I don't think we need to address this in this PR, since there is no regression for that use case (since this is modifying how variables are rendered, not how to render variables that haven't been defined).

So I removed the TODO here.

yorinasub17: Also see https://github.com/gruntwork-io/boilerplate/issues/93

Issue #96, opened 2022-03-01
Implement version enforcement feature


yorinasub17:
**Describe the solution you'd like**
Add the ability to specify a version string to enforce that a particular user invoking a boilerplate template has the specified required version installed.

Something like:

`boilerplate.yml`
```yaml
required_version: "~> 4.2"
```

**Describe alternatives you've considered**
Right now, you have to manually enforce boilerplate versions.

PR #98, opened 2022-03-04
Version enforcement

yorinasub17:
## Description

This adds the ability to require specific versions of boilerplate for processing the config.

## TODOs

- [x] Update the docs.

## Related Issues

Fixes https://github.com/gruntwork-io/boilerplate/issues/96

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -210,10 +210,7 @@ func BoilerplateConfigPath(templateFolder string) string {
 // against the current version of the CLI.
 func EnforceRequiredVersion(boilerplateConfig *BoilerplateConfig) error {
 	// Base case: if required_version is not set, then there is no version to enforce.
-	if boilerplateConfig == nil {
-		return nil
-	}
-	if boilerplateConfig.RequiredVersion == nil {
+	if boilerplateConfig == nil || boilerplateConfig.RequiredVersion == nil {
 		return nil
 	}
 	constraint := *boilerplateConfig.RequiredVersion


config/config.go
```diff
@@ -196,10 +206,53 @@ func BoilerplateConfigPath(templateFolder string) string {
...
+// EnforceRequiredVersion enforces any required_version string that is configured on the boilerplate config by checking
+// against the current version of the CLI.
+func EnforceRequiredVersion(boilerplateConfig *BoilerplateConfig) error {
+	// Base case: if required_version is not set, then there is no version to enforce.
+	if boilerplateConfig == nil {
+		return nil
+	}
+	if boilerplateConfig.RequiredVersion == nil {
+		return nil
+	}
```
rhoboat: This can't be simplified? If you `||` these, then won't it fail quickly if the first one is false?

.circleci/config.yml
```diff
@@ -33,7 +33,10 @@ jobs:
           name: run tests
           command: |
             mkdir -p /tmp/logs/results
-            run-go-tests --path . --timeout 1h | tee /tmp/logs/all.log
+            # Some unit tests depend on a version being set, so here we pass in the ldflags to set the version to the
+            # expected test string.
+            run-go-tests --path . --timeout 1h --packages "-ldflags '-X github.com/gruntwork-io/go-commons/version.Version=v1.33.7' ./..." \
```
rhoboat, line 38: I'm curious.. is this now the version of go-commons, which is different from the version of boilerplate? I'm missing some relationship here. I assumed go-commons is a library used by boilerplate and its versioning is handled separately.

integration-tests/required_version_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+const (
+	testVersion = "v1.33.7"
+)
+
+func TestRequiredVersionMatchCase(t *testing.T) {
+	t.Parallel()
+
+	// Make sure that the test is run with the ld flags setting version to our expected test version.
+	require.Equal(t, testVersion, version.GetVersion())
```
rhoboat, line 24: Does this mean we must continually bump the `testVersion` on line 17 to equal whatever the current version is? Or something I'm not understanding.

.circleci/config.yml
yorinasub17: Are you referring to the version string I am passing in here? This version is just a random test string that I set so that we can test the functionality around required versions.

If I don't set this, then the version of the CLI will be `""`, which makes it hard to test the enforcement of required versions.

integration-tests/required_version_test.go
yorinasub17: No the point is to set to a specific, but arbitrary version so that we can unit test this functionality against various `required_version` strings in a consistent manner.

I'm doing this primarily because we can't really set this to the "current version" because PRs won't have an actual version associated with it.

config/config.go
yorinasub17: Done: afb9df44fcf56c5a3539aa35b234eb631aa1bf6b

.circleci/config.yml
rhoboat: Ah, I get it now. This is a value passed into the test so that the `testVersion` can be evaluated against it.

I am still wondering why this PR is about specifying a boilerplate version when this looks like it's specifying a go-commons version.

rhoboat: Also I see the 1337 reference. 🥇

integration-tests/required_version_test.go
rhoboat: Yup, Sorry I missed the point, but get it now. Makes sense.

.circleci/config.yml
yorinasub17: So the `-X` flag for `-ldflags` is a way to inject a value for a global variable at build time. This allows you to dynamically control values in the global namespace as you are compiling.

In essence, what is going on here is that, just for the test, we are setting the `Version` var in the namespace `github.com/gruntwork-io/go-commons/version` to the value `v1.33.7` without hardcoding that value in the source code.

This value is later set to the release tag when we are ready to publish the binaries using the same mechanism, here: https://github.com/gruntwork-io/boilerplate/pull/98/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R69

I know it looks confusing because it is setting the `Version` global var in the `version` package, but this is just how we designed the `version` library in `go-commons`. You can read the comments there for more info: https://github.com/gruntwork-io/go-commons/blob/master/version/version.go

PR #102, opened 2022-04-27
Implement validation system and user prompt enhancements

zackproser:
## Description

These changes introduce a few enhancements that we're planning to work into an updated `gruntwork` CLI experience. Namely: 

**Validations** - users can specify a limited subset of known validations in their boilerplate.yml to have those validations run in real-time against their inputs when boilerplate prompts the end user for an input. 

boilerplate-validations

**Enum types now present multi-select prompts** to the user with the specified options
boilerplate-enum

## TODOs

- [ ] Ensure the branch is named correctly with the issue number. e.g: `feature/new-vpc-endpoints-955` or `bug/missing-count-param-434`.
- [ ] Update the docs.
- [ ] Keep the changes backward compatible where possible.
- [ ] Run the pre-commit checks successfully.
- [ ] Run the relevant tests successfully.

### DIFF
--- a/variables/yaml_helpers.go
+++ b/variables/yaml_helpers.go
@@ -256,21 +256,20 @@ func ConvertValidationStringtoRules(ruleString string) ([]CustomValidationRule, 
 			}
 			validationRules = append(validationRules, cvr)
 		}
-
 	}
 
 	return validationRules, nil
 }
 
-// Given a map of key:value pairs read from a Boilerplate YAML config file of the format:
+// Given a list of validations read from a Boilerplate YAML config file of the format:
 //
 // validations:
-//   - foo
-//   - bar
-//   - baz
+//   - url
+//   - email
+//   - length-5-20
 //
-// This method takes looks up the options object in the map and unmarshals the data inside of it it into a list of
-// strings.
+// This method looks up the validations specified in the map and applies them to the specified fields so that users prompted for input
+// get real-time feedback on the validity of their entries
 func unmarshalValidationsField(fields map[string]interface{}, context string, variableType BoilerplateType) ([]CustomValidationRule, error) {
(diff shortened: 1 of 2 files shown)


variables/yaml_helpers.go
```diff
@@ -175,6 +178,111 @@ func unmarshalOptionsField(fields map[string]interface{}, context string, variab
...
+		validatorsToReturn = append(validatorsToReturn, rule.Validator)
+	}
+	return validatorsToReturn
+}
+
+func (c CustomValidationRule) DescriptionText() string {
+	return c.Message
+}
+
+func ConvertValidationStringtoRules(ruleString string) ([]CustomValidationRule, error) {
```
zackproser, line 200: This is kind of a bummer. What I really wanted was a way parse the user-supplied strings, such as `is.URL` or `validation.Required` and just feed those directly to the validation library, but I've not yet figured out how to accomplish that.

yorinasub17: Ah that's because go is a static compiled language, so it doesn't have a way to do dynamic injection.

The best alternative approach to do what you want would be to use a secondary dynamic language as the runtime that go calls out to. For example, you can have the validation logic be written in lua and use [gopher-lua](https://github.com/yuin/gopher-lua) to run it.

zackproser: Ah thanks, Yori. I thought perhaps I could abuse reflection somehow to get around this :(

yorinasub17: Reflection is a read-only operation, which works/makes sense because you are reading the state of the go code, but not compiling/injecting in new code (which requires calling out to the go compiler and dynamically morphing the binary - physically impossible - OR using a go interpreter).

PR #105, opened 2022-04-29
Implement variable-level validations for gruntwork cli

zackproser:
Closes https://github.com/gruntwork-io/company/issues/460

These changes implement variable-level validations that can be defined in `boilerplate.yml`. For example: 

```
 - name: CompanyName
    description: Your company name is used as a prefix when naming resources like your S3 Buckets and Load Balancers. It must remain short, so that it does not exceed AWS name length limits.
    type: string
    validations:
      - required
      - length-5-22
      - alphanumeric
```
Will result in the user being presented with a prompt that will only accept a value that passes all three of the defined validation rules: 

* The variable can't be empty
* It must have a length between 5 and 22 characters
* It can only contain alphanumeric characters

Note that, while we're planning on using this branch in the updated `gruntwork` CLI experience we're shipping as part of the Spring 2022 Ref Arch Moratorium, we're **not** necessarily planning on merging this into boilerplate permanently, as-written. 

Ultimately, It's  important that we get a working CLI experience with validations that can be completed quickly by our users, but we don't want to overly invest in this interim solution, either, because it only needs to bridge the current gap until our next offering is ready. 

## TODOs

- [x] Update the docs.

### DIFF
--- a/config/get_variables.go
+++ b/config/get_variables.go
@@ -79,7 +79,7 @@ func GetVariables(opts *options.BoilerplateOptions, boilerplateConfig, rootBoile
 	}
 
 	// Create a slice of KeyOrderPairs, which we'll be able to sort according to order value
-	keyAndOrderPairs := make([]KeyAndOrderPair, len(variablesInConfig))
+	keyAndOrderPairs := []KeyAndOrderPair{}
 
 	// Pair the keys for each variable to its user-defined presentation order
 	for key, variable := range variablesInConfig {


config/get_variables.go
```diff
@@ -207,15 +196,71 @@ func getVariableFromUser(variable variables.Variable, opts *options.BoilerplateO
...
+		if hasValidationErrs {
+			ie := variables.InvalidEntries{
+				Issues: []variables.ValidationIssue{
+					{Value: value,
+						ValidationMap: m},
+				},
+			}
+			return getVariableFromUser(variable, opts, ie)
+		}
+	}
```
yorinasub17: You probably don't have to guard with `if len(variable.Validations()) > 0 {` because the `range` naturally handles that for you.

As in, when length of `variable.Validations()` is `0`, the range for loop won't do anything, leaving `hasValidationErrs` as `false`.

config/get_variables.go
```diff
@@ -207,15 +196,71 @@ func getVariableFromUser(variable variables.Variable, opts *options.BoilerplateO
...
+func renderValidationErrors(val interface{}, m map[string]bool) {
+	util.ClearTerminal()
+	pterm.Warning.WithPrefix(pterm.Prefix{Text: "Invalid entry"}).Println(val)
+	for k, v := range m {
+		if v == false {
+			pterm.Error.Println(k)
+		} else if v == true {
+			pterm.Success.Println(k)
+		}
+	}
```
yorinasub17: NIT: I think the following feels more natural for a simple if-else clause on a bool:

```suggestion
		if v {
			pterm.Success.Println(k)
		} else {
			pterm.Error.Println(k)
		}
	}
```

util/prompt.go
```diff
@@ -41,3 +41,7 @@ func PromptUserForYesNo(prompt string) (bool, error) {
 		return false, nil
 	}
 }
+
+func ClearTerminal() {
```
yorinasub17, line 47: NIT: Add a reference link. Also, this probably only works on unix and not on windows. Might be better to add a `prompt_unix.go` file with build rules for `linux | darwin` and add a separate `windows` file that does nothing.

variables/yaml_helpers.go
```diff
@@ -175,6 +178,111 @@ func unmarshalOptionsField(fields map[string]interface{}, context string, variab
...
+		validatorsToReturn = append(validatorsToReturn, rule.Validator)
+	}
+	return validatorsToReturn
+}
+
+func (c CustomValidationRule) DescriptionText() string {
+	return c.Message
+}
+
+func ConvertValidationStringtoRules(ruleString string) ([]CustomValidationRule, error) {
```
yorinasub17, line 217: NIT:
```suggestion
func ConvertValidationStringToRules(ruleString string) ([]CustomValidationRule, error) {
```

yorinasub17, line 217: Also, add function docs, including a list of what we intend to support, and a "why" comment indicating that this is a short term solution for satisfying limited validation rules for our needs.

config/get_variables.go
zackproser: Makes sense - removed guard in 8ca0fbd

util/prompt.go
zackproser: Roger - poked around and decided to give this library a try for the time being: https://github.com/inancgumus/screen

zackproser: Done in 332b079

variables/yaml_helpers.go
zackproser: Done in da0edf1

variables/yaml_helpers.go
```diff
@@ -324,6 +324,21 @@ func unmarshalTypeField(fields map[string]interface{}, context string) (Boilerpl
 	return BOILERPLATE_TYPE_DEFAULT, nil
 }
 
+func unmarshalOrderField(fields map[string]interface{}, fieldName string, requiredField bool) (int, error) {
```
yorinasub17: NIT: this should be called `unmarshalIntField`

zackproser: Nice, thank you for reviewing my branch :smile: ! Fixed this in e220d95

PR #134, opened 2023-06-19
Add for_each

brikis98:
## Description

This PR adds `for_each` and `for_each_reference` support to the `dependency` block of `boilerplate`. If you pass in a list to these attributes, instead of creating the `dependency` once, `boilerpalte` will loop over each item in the `for_each` list, and run the dependency once for each of those items. You can access the current item in the loop using `__each__`.

The main use case for this is to generate sub-folders and files dynamically: e.g., in https://github.com/gruntwork-io/prototypes/pull/361, I use this to generate environments (dev, stage, prod) dynamically, based on which environments the user asked for.

Bonus fix: remove Rho from CODEOWNERS; GitHub was showing an error about that.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

### DIFF
--- a/README.md
+++ b/README.md
@@ -365,7 +365,7 @@ executing the current one. Each dependency may contain the following keys:
 * `for_each`: If you set this to a list of values, `boilerplate` will loop over each value, and render this dependency
   once for each value. This allows you to dynamically render a dependency multiple times based on user input. The 
   current value in the loop will be available as the variable `__each__`, available to both your Go templating and
-  in other `dependencies` params: e.g., you could reference `{{ __each__ }}` in `output-folder` to render to each 
+  in other `dependencies` params: e.g., you could reference `{{ .__each__ }}` in `output-folder` to render to each 
   iteration to a different folder.
 * `for_each_reference`: The name of another variable whose value should be used as the `for_each` value.  
 


README.md
```diff
@@ -360,6 +362,12 @@ executing the current one. Each dependency may contain the following keys:
     - Defaults set on dependency variables (the `variables` field of the dependency).
     - Defaults set on root variables.
     - Defaults set within the dependency boilerplate config.
+* `for_each`: If you set this to a list of values, `boilerplate` will loop over each value, and render this dependency
+  once for each value. This allows you to dynamically render a dependency multiple times based on user input. The 
+  current value in the loop will be available as the variable `__each__`, available to both your Go templating and
+  in other `dependencies` params: e.g., you could reference `{{ __each__ }}` in `output-folder` to render to each
```
zackproser: In your example here, you use the `{{ __each__}}` reference - but below, each is prefixed with a period: `{{ .__each__ }}`. What's the difference? If there is one, is it worth a separate mention in docs / examples?

brikis98: Ah, whoops, you're right, updated to `.__each__`. Go template needs the `.` to look data up in maps and structs. https://github.com/gruntwork-io/boilerplate/pull/134/commits/43bab7cc1494c570adebc3ba5d65e0e0523aa372

PR #147, opened 2023-12-07
Fix slice parsing

brikis98:
## Description

As per https://github.com/urfave/cli/issues/1134, `urfave/cli` allows you to pass in "slices" (CLI args that can have multiple values) via either:

1. Passing the arg multiple times: e.g., `--var first --var second --var third`.
2. Passing multiple values to the arg with a separator (default: comma): e.g., `--var first,second,third`.

This second behavior is problematic for boilerplate, as we allow you to pass JSON and YAML into `--var` values, and JSON and YAML can contain commas. So these JSON and YAML values were being broken up in the middle, which made them invalid... And the error message we got was super unhelpful:

```
yaml: line 1: did not find expected ',' or '}'
```

This PR disables the second type of slice parsing. I also added a regression test that would fail with the unhelpful error above before the PR and passes now after the PR.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

### DIFF
--- a/cli/boilerplate_cli.go
+++ b/cli/boilerplate_cli.go
@@ -94,6 +94,13 @@ func CreateBoilerplateCli() *cli.App {
 		},
 	}
 
+	// We pass JSON/YAML content to various CLI flags, such as --var, and this JSON/YAML content may contain commas or
+	// other separators urfave/cli would treat as a slice separator, and would therefore break the value into multiple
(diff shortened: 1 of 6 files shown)


yhakbar (approved):
Looks good. It's pretty convenient that Golang CLI tooling makes enabling/disabling that behavior so easy.

brikis98:
Thanks for the review! Merging now.

PR #176, opened 2024-03-26
Improving binary/text files identification

### LINKED ISSUE #126: Do a better job of disambiguating between text and binary
**Describe the bug**
`boilerplate` has to distinguish between text files, which are passed through the Go templating engine, and binary files, which are copied without any changes. It tries to do this using the `file` command, if it's available on the computer: 

https://github.com/gruntwork-io/boilerplate/blob/d323184f59f1eb937fb9b6f7fb8ffb8ba1fc9fb5/util/file.go#L41-L51 

On computers without `file`, it falls back to Go's `http.DetectContentType`, which works well for the most part, but sometimes, it fails: for example, it fails on `.hcl` files, treating them as binary! As a result, the Go templating in the `.hcl` files doesn't get processed.

**To Reproduce**
Run `boilerplate` on a system that doesn't have `file` installed and have it process a `.hcl` file in the `--template-folder`.

**Expected behavior**
`boilerplate` should:

1. Ideal solution: detect all text vs binary files correctly. Is there a better Go library we can use? 
2. Alternative: at the very least, we should improve logging:
    1. Log if we think the file is text (and templating will be processed) or binary (so it's copied unchanged), as that will help users understand what's going on.
    2. Log if `file` is not installed and we're falling back to `http.DetectContentType`.

**Additional context**
Originally reported in https://gruntwork-io.slack.com/archives/CJ39EV0KW/p1669017123927079.

denis256:
## Description

Improving binary/text files identification

Fixes #126.

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -48,6 +48,7 @@ require (
 	github.com/google/uuid v1.2.0 // indirect
 	github.com/googleapis/gax-go/v2 v2.0.5 // indirect
 	github.com/gookit/color v1.5.0 // indirect
+	github.com/h2non/filetype v1.1.3 // indirect
 	github.com/hashicorp/errwrap v1.1.0 // indirect
 	github.com/hashicorp/go-cleanhttp v0.5.2 // indirect
 	github.com/hashicorp/go-safetemp v1.0.0 // indirect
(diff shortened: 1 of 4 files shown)


brikis98 (approved):
Oh, nice find! LGTM.

PR #189, opened 2024-08-11
switch logging to slog

### LINKED ISSUE #82: Implement logger with logging levels
The current logger in `boilerplate` only supports `Printf` with no logging levels. We should replace it with a logger that supports logging levels so that user can control the verbosity of the logs.

jlambert121:
## Description

Fixes #82.

Currently all logging is done via `utils.Logger` which just outputs logging to `stdout`.  It would be nice to have more control over logging by silencing logging when not needed or injecting a separate logger (compatible with `*slog.Logger`). 

There is a lot of lines of change here, but it's just creating a `slog.Handler` that prefixes `[boilerplate]` to log lines to maintain the existing look and passing an instance of the logger throughout.  `go` is not my primary language though, any feedback is greatly appreciated!

NOTE: This does bump the minimum go version to `1.21`.

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

Updated logging to use `slog` library.

### Migration Guide

None

### DIFF
--- a/options/options.go
+++ b/options/options.go
@@ -2,6 +2,7 @@ package options
 
 import (
 	"fmt"
+	"log/slog"
 
 	"github.com/urfave/cli/v2"
 
@@ -20,6 +21,7 @@ const OptMissingConfigAction = "missing-config-action"
 const OptDisableHooks = "disable-hooks"
 const OptDisableShell = "disable-shell"
 const OptDisableDependencyPrompt = "disable-dependency-prompt"
+const OptSilent = "silent"
 
 // The command-line options for the boilerplate app
 type BoilerplateOptions struct {
@@ -36,6 +38,8 @@ type BoilerplateOptions struct {
 	DisableHooks            bool
 	DisableShell            bool
 	DisableDependencyPrompt bool
+	Silent                  bool
+	Logger                  *slog.Logger
 }
 
 // Validate that the options have reasonable values and return an error if they don't
@@ -96,6 +100,7 @@ func ParseOptions(cliContext *cli.Context) (*BoilerplateOptions, error) {
 		DisableHooks:            cliContext.Bool(OptDisableHooks),
 		DisableShell:            cliContext.Bool(OptDisableShell),
 		DisableDependencyPrompt: cliContext.Bool(OptDisableDependencyPrompt),
+		Silent:                  cliContext.Bool(OptSilent),
 	}
 
 	if err := options.Validate(); err != nil {
--- a/util/prefixer/prefixer.go
+++ b/util/prefixer/prefixer.go
@@ -1,0 +1,50 @@
+package prefixer
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"log/slog"
+	"os"
+	"sync"
+)
+
+type Handler struct {
+	h      slog.Handler
+	m      *sync.Mutex
+	writer io.Writer
+}
+
+func (h *Handler) Enabled(ctx context.Context, level slog.Level) bool {
+	return h.h.Enabled(ctx, level)
+}
+
+func (h *Handler) WithAttrs(attrs []slog.Attr) slog.Handler {
+	return &Handler{h: h.h.WithAttrs(attrs), m: h.m, writer: h.writer}
+}
+
+func (h *Handler) WithGroup(name string) slog.Handler {
+	return &Handler{h: h.h.WithGroup(name), m: h.m, writer: h.writer}
+}
+
+func (h *Handler) Handle(ctx context.Context, r slog.Record) error {
+	msg := slog.StringValue(r.Message).String()
+	if msg != "" {
+		_, err := io.WriteString(h.writer, fmt.Sprintf("[boilerplate] %s\n", msg))
+		return err
+	}
+
+	return nil
+}
+
+func New() *Handler {
+	buf := &bytes.Buffer{}
+	handler := &Handler{
+		h:      slog.NewTextHandler(buf, &slog.HandlerOptions{}),
+		m:      &sync.Mutex{},
+		writer: os.Stdout,
+	}
+
+	return handler
+}
--- a/util/shell.go
+++ b/util/shell.go
@@ -3,14 +3,13 @@ package util
 import (
 	"os"
 	"os/exec"
-	"strings"
 
 	"github.com/gruntwork-io/boilerplate/errors"
 )
 
 // Run the given shell command with the given environment variables and arguments in the given working directory
 func RunShellCommandAndGetOutput(workingDir string, envVars []string, command string, args ...string) (string, error) {
-	Logger.Printf("Running command: %s %s", command, strings.Join(args, " "))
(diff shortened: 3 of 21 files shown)


util/shell.go
```diff
@@ -1,16 +1,20 @@
...
 	"strings"
 
 	"github.com/gruntwork-io/boilerplate/errors"
+	"github.com/gruntwork-io/boilerplate/util/prefixer"
 )
 
 // Run the given shell command with the given environment variables and arguments in the given working directory
 func RunShellCommandAndGetOutput(workingDir string, envVars []string, command string, args ...string) (string, error) {
-	Logger.Printf("Running command: %s %s", command, strings.Join(args, " "))
+	logger := slog.New(prefixer.New())
```
Resonance1584, line 16: We'll need to use the same instance of the logger here, and below, to correctly respect the silent option.

util/prefixer/prefixer.go
```diff
@@ -0,0 +1,50 @@
...
+
+	return nil
+}
+
+func New() *Handler {
+	buf := &bytes.Buffer{}
+	handler := &Handler{
+		h:      slog.NewTextHandler(buf, &slog.HandlerOptions{}),
+		m:      &sync.Mutex{},
+		writer: os.Stdout,
```
Resonance1584, line 46: Lets use os.Stderr here, it's more typical for logs to be sent to STDERR whereas machine readable program output should go to STDOUT

util/prefixer/prefixer.go
```diff
@@ -0,0 +1,50 @@
...
+func (h *Handler) WithAttrs(attrs []slog.Attr) slog.Handler {
+	return &Handler{h: h.h.WithAttrs(attrs), m: h.m, writer: h.writer}
+}
+
+func (h *Handler) WithGroup(name string) slog.Handler {
+	return &Handler{h: h.h.WithGroup(name), m: h.m, writer: h.writer}
+}
+
+func (h *Handler) Handle(ctx context.Context, r slog.Record) error {
+	msg := slog.StringValue(r.Message).String()
```
Resonance1584, line 32: This feels weird because we're not using any of the attributes or grouping from the slog package - but this is still a net improvement allowing us to use --silent so I'm happy for this to go in as long as there's a TODO comment explaining it. Something like:
```golang
// TODO: Handle attributes and groups
_, err := fmt.Fprintf(h.writer, "[boilerplate] %s\n", r.Message)
return err
```

util/prefixer/prefixer.go
```diff
@@ -0,0 +1,50 @@
...
+	"fmt"
+	"io"
+	"log/slog"
+	"os"
+	"sync"
+)
+
+type Handler struct {
+	h      slog.Handler
+	m      *sync.Mutex
```
Resonance1584, line 15: Unless I'm reading this wrong, this mutex doesn't appear to be used and should be removed.

options/options.go
```diff
@@ -36,6 +38,8 @@ type BoilerplateOptions struct {
 	DisableHooks            bool
 	DisableShell            bool
 	DisableDependencyPrompt bool
+	Silent                  bool
+	Logger                  *slog.Logger
```
Resonance1584, line 42: This doesn't feel right, I think we should either use a global slog.Default() or pass the logger instance as a function argument everywhere.

Resonance1584 (changes requested):
Hi @jlambert121 , thanks for your contribution! This seems like a good improvement to Boilerplate.

I don't think `BoilerplateOptions` is the correct place for our Logger instance, that struct is specific to parsing configuration. There's two ways forward here:

1. Least work:

Use [slog.SetDefault](https://pkg.go.dev/log/slog#SetDefault) after parsing the --silent option, then use the slog.Info etc methods at log callsites. This is pretty similar to the way the existing logger is utilized, and allows us to use the correct logger in `util/shell.go` without modifying any function signatures.

2. More work:

Pass the instance of *slog.Logger everywhere as a function argument. Typically I make the first two arguments of my functions context.Context and Logger. This allows different code paths to construct new logger instances with different key val attributes. There's not much value in doing that here as we're not using log attributes, so its more of a style preference to avoid the global.

I'd suggest going with the first (least work) option for now and we can revisit in the future if we want to start using log attributes :)

jlambert121:
@Resonance1584 thanks for taking a look at this!

> I don't think BoilerplateOptions is the correct place for our Logger instance

Agree, was trying to limit the signature changes until there was some feedback though.  In order to have an optional passed in logger (we're utilizing boilerplate inside another application), would you be opposed to changing the signature on `ProcessTemplate` to take a logger as well?

Resonance1584:
> @Resonance1584 thanks for taking a look at this!
> 
> > I don't think BoilerplateOptions is the correct place for our Logger instance
> 
> Agree, was trying to limit the signature changes until there was some feedback though. In order to have an optional passed in logger (we're utilizing boilerplate inside another application), would you be opposed to changing the signature on `ProcessTemplate` to take a logger as well?

@jlambert121 I actually hadn't considered the use of this project as a library (thanks for raising that use case!) - which makes signature changes on exported functions problematic.

What we could do here is add a new exported function for your use case, and modify the existing function so that they go through the same code path without breaking compatibility.

```golang
ProcessTemplate(options, rootOpts *options.BoilerplateOptions, thisDep variables.Dependency) error {
  var l *slog.Logger
  if options.Silent {
    l = slog.New(slog.NewTextHandler(io.Discard, nil))
  } else {
    l = slog.New(prefixer.New())
  }
  return ProcessTemplateWithLogger(l, options, rootOpts, thisDep)
}

ProcessTemplateWithLogger(l *slog.Logger, options, rootOpts *options.BoilerplateOptions, thisDep variables.Dependency) error {
  // .. previous body of ProcessTemplate goes here
}
```

Would that support your use case?

Unfortunately we also won't be able to modify `DownloadTemplatesToTemporaryFolder` or `RunShellCommand*` without breaking compatibility, which I think means we would need to continue using a global like `util/logger.go`. Using `slog.SetDefault` would be really intuitive for external applications.

jlambert121:
> @jlambert121 I actually hadn't considered the use of this project as a library (thanks for raising that use case!) - which makes signature changes on exported functions problematic.
> 
> What we could do here is add a new exported function for your use case, and modify the existing function so that they go through the same code path without breaking compatibility.
> 
>
> Would that support your use case?
> 
> Unfortunately we also won't be able to modify `DownloadTemplatesToTemporaryFolder` or `RunShellCommand*` without breaking compatibility, which I think means we would need to continue using a global like `util/logger.go`. Using `slog.SetDefault` would be really intuitive for external applications.

Yeah, this seems like probably the best option - I'll see about refactoring with this.  Yeah, `DownloadTemplatesToTemporaryFolder` and `RunShellCommand` are a bit more problematic.

pseudomorph:
What is the issue with signature changes here? If it's a matter of breaking changes being introduced, I believe there are some changes which are slated as such for Boilerplate in the near future (https://github.com/gruntwork-io/terragrunt/pull/4540#issuecomment-3114670501)

If that's the only option, perhaps these two matters can be combined into a single breaking release? Better logging would be great.

PR #219, opened 2025-04-16
fix: processing dependency variable, feat: render for_each_reference value

### LINKED ISSUE #215: Nested dependencies __each__ value is not as expected
**Describe the bug**
When I use nested dependencies with for_eaches, the inner dependency takes the `__each__` value from the outer dependency

I found the the loop which overrides the `__each__` value.
https://github.com/gruntwork-io/boilerplate/blob/main/templates/template_processor.go#L400

If I comment that loop out boilerplate works perfectly. 

To me it seems like the dependency name is not listed before the variable name in `opts.Vars`, so the statement is always true and overrides the `__each__` with the original value.

**To Reproduce**
I created a test config for this purpose, I'm uploading it.
Running command:
```shell
boilerplate --template-url blueprints --output-folder test_output --var-file auto_vars.yml --non-interactive
```

**Expected behavior**
Inner dependency uses its own `__each__` value.

**Other Question**
There's one more thing I can't decide if it's a bug or an improvement.
So I ask first.
I'd like to use go templating in `for_each_reference`, is it possible?

The config contains a use case:

`templates/dep1_template/boilerplate.yml`
```yml
for_each_reference: '{{ $region_snake := .region | snakeCase }}{{ index .deployments $region_snake "template" }}'
```

This template works perfectly if I use it as a variable value, so maybe the `for_each_reference` doesn't render the template.

Should I create a separate issue of it?

[boilerplate_test.zip](https://github.com/user-attachments/files/19703818/boilerplate_test.zip)

tiviuray:
## Description

Fixes #215.

## TODOs

- [ ] Update the docs.

## Release Notes

Updated [X].
fix: dependency variable processing to evaluate template strings during variable inheritance.
feat: add capability for for_each_reference to process go template.


tiviuray:
Sure, there is a zip file in the issue I opened for it

tiviuray:
@Resonance1584 I checked the tests, but I'm a bit confused about what this test is supposed to test? The functions used are already tested, basically rendering, unmarshalling and map merging.

So, please be so kind and give me some help.

Resonance1584:
Hi @tiviuray

Please add the test case from @pseudomorph (thankyou!), then push this branch to this repository and update the PR to reference the new branch - that will allow me to run our test suite against it

Resonance1584:
@tiviuray thanks! Ignore my note about pushing to this repo - it's on me to kick off the tests. I'll do so and follow up in the other PR

PR #225, opened 2025-06-21
[WIP] feat: Add optional function to return manifest of generated files

### LINKED ISSUE #224: Add list of generated files as optional output
**Describe the solution you'd like**
When invoking Boilerplate it should optionally provide an output (list) of all the files it created.

For instance, a cli flag could be added: `--file-manifest`

**Additional context**
This would be a feature to help address [this terragrunt issue](https://github.com/gruntwork-io/terragrunt/issues/3657). Currently Terragrunt runs an hclfmt on the entire output/working directory to ensure all generated files are properly formatted. This is problematic due to reasons described in that ticket. In order to address the issue Terragrunt needs to know which files were generated. Terragrunt itself however has no direct way of figuring out which files were created by Boilerplate. The most logical way to proceed is to add the above functionality, so that terragrunt can consume the Boilerplate output and then format the files listed.

tgeijg:
## Description

Fixes #224.

Adding a function to return the manifest of files that boilerplate generated. This would be the most clean way to work towards resolving [this terragrunt issue](https://github.com/gruntwork-io/terragrunt/issues/3657).

I tried keeping things backwards compatible. The cleaner option would be to update the `processTemplate` function to retun an output, but since that would change the function signature it would be a breaking change. Happy to make that change if desired.

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

Added function which returns a manifest of all generated files.


templates/template_processor.go
```diff
@@ -22,12 +24,23 @@ import (
...
+	return processTemplateInternal(options, rootOpts, thisDep, true)
+}
+
 // Process the boilerplate template specified in the given options and use the existing variables. This function will
 // download remote templates to a temporary working directory, which is cleaned up at the end of the function. This
 // function will load any missing variables (either from command line options or by prompting the user), execute all the
 // dependent boilerplate templates, and then execute this template. Note that we pass in rootOptions so that template
 // dependencies can inspect properties of the root template.
-func ProcessTemplate(options, rootOpts *options.BoilerplateOptions, thisDep variables.Dependency) error {
+func processTemplateInternal(options, rootOpts *options.BoilerplateOptions, thisDep variables.Dependency, returnManifest bool) (*manifest.Manifest, error) {
```
yhakbar: There's a more idiomatic way to do this, and I think it's worth breaking backwards compatibility to achieve it.

I'd recommend creating a `NewTemplateProcessor` function that returns a template processor, accepting different options to configure it, then having a `Process()` method to do what it needs to do.

This is an example of that:
https://github.com/gruntwork-io/terragrunt/blob/main/internal/discovery/discovery.go#L107

It also allows us to get out of passing around global options like that.

I'm not the lead maintainer of this project, so I'll solicit some feedback from other engineers at Gruntwork. I recommend waiting for their input before making any changes to this PR.

yhakbar (changes requested):
If we're adding the ability to generate a manifest like this to the project, it should be an accessible option from the CLI as well. Regardless of how we tackle adjusting the backwards compatibility of the library, I think we should also look to expose that functionality in the `boilerplate` CLI.

yhakbar:
Also, I think using the CLI flag should write the manifest to disk, with some versioned schema that can be used by Boilerplate to do intelligent things like clean up files that were previously generated, but are no longer part of the template, etc.

templates/template_processor.go
yhakbar: We've discussed this internally, and I can bless the decision to engage in a breaking change to make this codepath more maintainable by changing the API.

We'll look to merge this in with another anticipated breaking change to require confirmation before running shells/hooks for users when in interactive mode.

tgeijg:
> Also, I think using the CLI flag should write the manifest to disk, with some versioned schema that can be used by Boilerplate to do intelligent things like clean up files that were previously generated, but are no longer part of the template, etc.

@yhakbar  just to make sure I understand this correctly: When we talk about a versioned manifest file, we're essentially saying that on function invocation we're going to check if a manifest file already exists in the working directory, and if it does we update it, creating a versioned file in the process? Or do we expect the CLI flag to take in a directory input to where the manifest file needs to go/be read from?

templates/template_processor.go
tgeijg: This has been updated. I just need some feedback as per below regarding the versioned manifest. Should be able to update early next week.

tgeijg:
> Also, I think using the CLI flag should write the manifest to disk, with some versioned schema that can be used by Boilerplate to do intelligent things like clean up files that were previously generated, but are no longer part of the template, etc.

@yhakbar Added some changes to how I imagine the manifest could/should work with versioning. Open to feedback/suggestions.

README.md
```diff
@@ -134,13 +134,14 @@ You can find older versions on the [Releases Page](https://github.com/gruntwork-
    you to run arbitrary scripts.
 1. **Cross-platform**: Boilerplate is easy to install (it's a standalone binary) and works on all major platforms (Mac,
    Linux, Windows).
+1. **Versioned file manifest**: Optionally generate a versioned JSON manifest that tracks all created files across multiple template generations. Each time boilerplate runs with `--output-manifest`, it adds a new version entry to the existing manifest file (or creates one if none exists), preserving the history of all previous generations for integration with other tools.
```
yhakbar, line 137: Why are we preserving all previous versions?

README.md
```diff
@@ -185,6 +186,7 @@ The `boilerplate` binary supports the following options:
 * `--no-shell`: If this flag is set, no `shell` helpers will execute. They will instead return the text "replace-me".
 * `--disable-dependency-prompt` (optional): Do not prompt for confirmation to include dependencies. Has the same effect as
   --non-interactive, without disabling variable prompts. Default: `false`.
+* `--output-manifest` (optional): Write a versioned JSON manifest of all generated files to `boilerplate-manifest.json` in the output directory. If a manifest file already exists, a new version entry will be added to track the current generation alongside previous ones.
```
yhakbar, line 189: Would a better name for this flag be `--manifest`?

We don't require following these rules [here](https://terragrunt.gruntwork.io/docs/reference/cli/rules/), but if we can, let's go with `--<system name>` when the name of the system can sufficiently capture what the flag is controlling. Other names could include `--manifest-generate` or `--manifest-file` if users can specify what they want the manifest file to be called, and specifying the name also determines that the file is generated.

cli/boilerplate_cli.go
```diff
@@ -96,6 +97,10 @@ func CreateBoilerplateCli() *cli.App {
 			Name:  options.OptDisableDependencyPrompt,
 			Usage: fmt.Sprintf("Do not prompt for confirmation to include dependencies. Has the same effect as --%s, without disabling variable prompts.", options.OptNonInteractive),
 		},
+		&cli.BoolFlag{
+			Name:  options.OptOutputManifest,
+			Usage: "Write a JSON list of all generated files to boilerplate-manifest.json in the current working directory.",
```
yhakbar, line 102: Did I request that this be in JSON instead of YAML? Seems at odds with the configuration file being in YAML, but it is a more convenient format in general. Maybe the format should be configurable. If we can get the UX to be similar to the [Run Report](https://terragrunt.gruntwork.io/docs/features/run-report/) that might be better, so there's an expectation of parity in how the two behave.

integration-tests/for_production_example_unix_test.go
```diff
@@ -19,9 +19,9 @@ func TestForProductionTerragruntArchitectureBoilerplateExample(t *testing.T) {
 
 	forProductionExamplePath := "../examples/for-production/terragrunt-architecture-catalog"
 
-	outputBasePath, err := ioutil.TempDir("", "boilerplate-for-production-output")
+	outputBasePath, err := os.MkdirTemp("", "boilerplate-for-production-output")
```
yhakbar, line 22: We should replace this with `t.TempDir` if we're going to replace this.

integration-tests/required_version_test.go
```diff
@@ -58,7 +57,7 @@ func TestRequiredVersionUnderTest(t *testing.T) {
 func runRequiredVersionExample(t *testing.T, templateFolder string) error {
 	app := cli.CreateBoilerplateCli()
 
-	outputPath, err := ioutil.TempDir("", "boilerplate-test-output-reqver")
+	outputPath, err := os.MkdirTemp("", "boilerplate-test-output-reqver")
```
yhakbar, line 60: We should replace this with `t.TempDir` if we're going to replace this.

manifest/manifest.go
```diff
@@ -0,0 +1,75 @@
...
+	Timestamp          string                 `json:"timestamp"`
+	TemplateURL        string                 `json:"template_url"`
+	BoilerplateVersion string                 `json:"boilerplate_version"`
+	Variables          map[string]interface{} `json:"variables"`
+	OutputDir          string                 `json:"output_dir"`
+	Files              []GeneratedFile        `json:"files"`
+}
+
+// VersionedManifest represents the complete versioned manifest
+type VersionedManifest struct {
```
yhakbar, line 28: This manifest file should include some information on the version of the manifest schema. That will be important if the schema changes.

manifest/manifest.go
```diff
@@ -0,0 +1,75 @@
...
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/gruntwork-io/go-commons/version"
+)
+
+// GeneratedFile represents a file that was generated by boilerplate
+type GeneratedFile struct {
```
yhakbar, line 13: Does it make sense to also include the checksums of the files generated?

manifest/manifest.go
```diff
@@ -0,0 +1,75 @@
...
+// VersionedManifest represents the complete versioned manifest
+type VersionedManifest struct {
+	LatestVersion string                   `json:"latest_version"`
+	Versions      map[string]ManifestEntry `json:"versions"`
+}
+
+
+
+// UpdateVersionedManifest updates the versioned manifest with a new entry
+func UpdateVersionedManifest(outputDir, templateURL string, variables map[string]interface{}, files []GeneratedFile) error {
```
yhakbar, line 36:
```suggestion
func UpdateVersionedManifest(outputDir, templateURL string, variables map[string]any, files []GeneratedFile) error {
```

manifest/manifest.go
```diff
@@ -0,0 +1,75 @@
...
+	manifestPath := filepath.Join(outputDir, "boilerplate-manifest.json")
+
+	// Read existing manifest or create new one
+	manifest := &VersionedManifest{
+		Versions: make(map[string]ManifestEntry),
+	}
+
+	if data, err := os.ReadFile(manifestPath); err == nil {
+		if err := json.Unmarshal(data, manifest); err != nil {
+			// If unmarshal fails, start with empty manifest
```
yhakbar, line 46: Wait, why? The file is corrupted, and we should throw an error, right?

templates/template_processor.go
```diff
@@ -714,14 +720,27 @@ func outPath(file string, opts *options.BoilerplateOptions, variables map[string
 }
 
 // Copy the given file, which is in options.TemplateFolder, to options.OutputFolder
-func copyFile(file string, opts *options.BoilerplateOptions, variables map[string]interface{}) error {
+func copyFile(file string, opts *options.BoilerplateOptions, variables map[string]interface{}, generatedFilePaths *[]string) error {
```
yhakbar, line 723: Why are we passing the pointer to `generatedFilePaths` here? Can't we manage this slice outside of this?

manifest/manifest_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestUpdateVersionedManifestMultipleVersions(t *testing.T) {
+	t.Parallel()
+
+	// Create temporary directory
+	tempDir, err := os.MkdirTemp("", "manifest-test")
```
yhakbar, line 18: Use `t.TempDir`

manifest/manifest_test.go
```diff
@@ -0,0 +1,77 @@
...
+	require.NoError(t, err)
+	defer os.RemoveAll(tempDir)
+
+	// First update
+	err = UpdateVersionedManifest(tempDir, "template1", map[string]interface{}{"var1": "value1"}, []GeneratedFile{{Path: "file1.txt"}})
+	require.NoError(t, err)
+
+	// Wait a moment to ensure different timestamps
+	time.Sleep(time.Second * 1)
```
yhakbar, line 27: Do we really need this?

manifest/manifest_test.go
```diff
@@ -0,0 +1,77 @@
...
+	latestEntry := manifest.Versions[manifest.LatestVersion]
+	assert.Equal(t, "template2", latestEntry.TemplateURL)
+	assert.Equal(t, map[string]interface{}{"var2": "value2"}, latestEntry.Variables)
+}
+
+func TestUpdateVersionedManifestInvalidExistingFile(t *testing.T) {
+	t.Parallel()
+
+	// Create temporary directory
+	tempDir, err := os.MkdirTemp("", "manifest-test")
```
yhakbar, line 56: Use `t.TempDir`

templates/template_processor_test.go
```diff
@@ -93,6 +93,97 @@ func TestCloneOptionsForDependency(t *testing.T) {
 	}
 }
 
+func TestProcessTemplateReturnsFiles(t *testing.T) {
+	t.Parallel()
+
+	// Create a temporary directory for the test
+	templateDir, err := os.MkdirTemp("", "template-test")
```
yhakbar, line 100: Use `t.TempDir`

templates/template_processor_test.go
```diff
@@ -93,6 +93,97 @@ func TestCloneOptionsForDependency(t *testing.T) {
...
+	content, err := os.ReadFile(filepath.Join(outputDir, "test.txt"))
+	require.NoError(t, err)
+	assert.Equal(t, "This is a test!", string(content))
+}
+
+func TestProcessTemplate(t *testing.T) {
+	t.Parallel()
+
+	// Create a temporary directory for the test
+	templateDir, err := os.MkdirTemp("", "template-test")
```
yhakbar, line 147: Use `t.TempDir`

templates/template_processor_test.go
```diff
@@ -93,6 +93,97 @@ func TestCloneOptionsForDependency(t *testing.T) {
...
+	require.NoError(t, err)
+	defer os.RemoveAll(templateDir)
+
+	// Create a temporary output directory
+	outputDir, err := os.MkdirTemp("", "output-test")
```
yhakbar, line 152: Use `t.Tempdir`

README.md
tgeijg: For clarity, are you suggesting we add multiple flags? 
`--manifest` returns the manifest, and
`--manifest-file <file_name>` writes the file to `<file_name>`? 

The only problem I see with the latter is that we rely on a consistent file name for the history/version feature. If someone were to enter a file name with a typo you would lose your version history, or it would be incomplete the next time you run the command with the correct file name.

README.md
tgeijg: I started there based on your previous comment: https://github.com/gruntwork-io/boilerplate/pull/225#issuecomment-3009391000

> I think using the CLI flag should write the manifest to disk, with some versioned schema that can be used by Boilerplate to do intelligent things like clean up files that were previously generated, but are no longer part of the template, etc

We could add some arbitrary version limit for sure, not sure what we'd consider reasonable though. I use boilerplate exclusively for scaffold, for which this isn't really a required feature. I just need the manifest to address the issue of terragrunt formatting the entire working directory. I can't really comment on other use cases of scaffold and/or what would be reasonable limits to version history.

cli/boilerplate_cli.go
tgeijg: That may have just been my default, this can be rewritten to yaml for sure. Will see how easy it is to make this configurable. Will have a look at Run Report.

PR #227, opened 2025-07-03
State management helpers

### LINKED ISSUE #226: Create `templateUrl` and `vars` helpers
This tool is awesome, but it is missing state management. Obviously, one would want to update the repo past initial generation.

I am currently trying to work around this with a bit of a `make`:

```make
.PHONY: boilerplate-update
boilerplate-update:
	boilerplate \
		--template-url {{ (.This.Options | toJson | fromJson).TemplateUrl }} \
		--output-folder . \
		--var-file ./boilerplate-vars.yml
```

And that `./boilerplate-vars.yml` it referencing? Part of the template:

```yaml
ModuleName: {{ .ModuleName }}
TofuVersion: {{ .TofuVersion }}
```

This way at any point since the initial generation, I can just run `make boilerplate-update` and it would update everything.

The problem is as you can see neither `templateUrl` nor `vars` are available currently. By inspecting `{{ . | toYaml }}`, I've found out that I can get access to template URL via `{{ (.This.Options | toJson | fromJson).TemplateUrl }}`. The to/from json is because the property is `.This.Options.templateurl` i.e. private and go-template doesn't allow me to access it. But since it had json struct tag, after marshaling and unmarshaling back it becomes `.This.Options.TemplateUrl`. 

For variables, I've found out when `--var-file` was specified then `{{ (.This.Options | toJson | fromJson).Vars | toYaml }}` may do the trick, but it is absent if the boilerplate was not ran with `--var-file` and answers were given interactively. So I have to define every var manually in my `boilerplate-vars.yml`.

Obviously this is such a hack. It is making use of a private API and it may break at any point. I wish the tool had state management baked in, but at the very least, there are some low hanging fruits that would help the users to achieve it with `make` in the meantime.

dee-kryvenko:
## Description

Fixes #226.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Added `templateUrl` and `boilerplateConfigVarsAll` helpers

### DIFF
--- a/render/template_helpers.go
+++ b/render/template_helpers.go
@@ -93,12 +93,14 @@ func CreateTemplateHelpers(templatePath string, opts *options.BoilerplateOptions
 
 		"templateIsDefined": wrapIsDefinedWithTemplate(tmpl),
 
-		"templateFolder":        func() (string, error) { return filepath.Abs(opts.TemplateFolder) },
-		"outputFolder":          func() (string, error) { return filepath.Abs(opts.OutputFolder) },
-		"relPath":               relPath,
-		"boilerplateConfigDeps": boilerplateConfigDeps(opts),
-		"boilerplateConfigVars": boilerplateConfigVars(opts),
-		"envWithDefault":        env,
+		"templateFolder":           func() (string, error) { return filepath.Abs(opts.TemplateFolder) },
+		"templateUrl":              func() string { return opts.TemplateUrl },
+		"outputFolder":             func() (string, error) { return filepath.Abs(opts.OutputFolder) },
+		"relPath":                  relPath,
+		"boilerplateConfigDeps":    boilerplateConfigDeps(opts),
+		"boilerplateConfigVars":    boilerplateConfigVars(opts),
+		"boilerplateConfigVarsAll": func() map[string]interface{} { return opts.Vars },
+		"envWithDefault":           env,
 
 		// DEPRECATIONS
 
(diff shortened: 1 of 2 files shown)


render/template_helpers.go
```diff
@@ -93,12 +93,14 @@ func CreateTemplateHelpers(templatePath string, opts *options.BoilerplateOptions
...
-		"boilerplateConfigDeps": boilerplateConfigDeps(opts),
-		"boilerplateConfigVars": boilerplateConfigVars(opts),
-		"envWithDefault":        env,
+		"templateFolder":           func() (string, error) { return filepath.Abs(opts.TemplateFolder) },
+		"templateUrl":              func() string { return opts.TemplateUrl },
+		"outputFolder":             func() (string, error) { return filepath.Abs(opts.OutputFolder) },
+		"relPath":                  relPath,
+		"boilerplateConfigDeps":    boilerplateConfigDeps(opts),
+		"boilerplateConfigVars":    boilerplateConfigVars(opts),
+		"boilerplateConfigVarsAll": func() map[string]interface{} { return opts.Vars },
```
dee-kryvenko: I am really tempted to call it just `vars` but I am not sure why `boilerplateConfigVars` called that way. Maybe there would be some clash. Let me know.

Resonance1584: Hi @dee-kryvenko , after reading through this I believe `boilerplateConfigVars` is named that way as it is intended to only contains vars defined via config.

I think using just `vars` here is more intuitive, lets go with that

PR #261, opened 2026-01-23
fix: address gocritic hugeParam and rangeValCopy lint errors

Resonance1584:
## Summary
- Updates method receivers and function parameters to use pointers for large structs (`Dependency`, `Hook`, `defaultVariable`)
- Converts range loops to use index-based access to avoid copying large structs
- Fixes lint failures caused by upstream terragrunt `.golangci.yml` enabling stricter `gocritic` checks

## Test plan
- [x] `make lint` passes with 0 issues
- [x] All unit tests pass (`go test ./config/... ./templates/... ./variables/...`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/config/get_variables.go
+++ b/config/get_variables.go
@@ -25,7 +25,7 @@ const MaxReferenceDepth = 20
 // GetVariables gets a value for each of the variables specified in boilerplateConfig, other than those already in existingVariables.
 // The value for a variable can come from the user (if the non-interactive option isn't set), the default value in the
 // config, or a command line option.
-func GetVariables(opts *options.BoilerplateOptions, boilerplateConfig, rootBoilerplateConfig *BoilerplateConfig, thisDep variables.Dependency) (map[string]any, error) {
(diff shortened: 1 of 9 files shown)


config/get_variables.go
```diff
@@ -44,8 +44,10 @@ func GetVariablesWithContext(ctx context.Context, opts *options.BoilerplateOptio
 
 	// Add a variable for all dependencies contained in the root config file. This will allow Golang template users
 	// to directly access these with an expression like "{{ .BoilerplateConfigDeps.foo.OutputFolder }}"
-	rootConfigDeps := map[string]variables.Dependency{}
-	for _, dep := range rootBoilerplateConfig.Dependencies {
+	rootConfigDeps := map[string]*variables.Dependency{}
```
yhakbar, line 47: If we're going to go through all this effort, we could be prealloc-ing this, right?

PR #263, opened 2026-01-23
Upgrade dependencies

Resonance1584:


.codespellrc
```diff
@@ -1,2 +1,4 @@
 [codespell]
-skip = template_helpers_test.go
+skip =
```
yhakbar, line 2: This is pretty weird.

PR #272, opened 2026-02-12
Add regex(pattern) validation type

josh-padnick:
This PR adds a `regex(pattern)` validation type that lets users validate variables against arbitrary regular expression patterns. For example:

```yaml
variables:
  - name: ProjectCode
    description: A project code like AB-1234 (two uppercase letters, dash, four digits)
    validations:
      - required
      - "regex(^[A-Z]{2}-\\d{4}$)"    # <-- new validation type
    default: GW-0001
```

Currently, users who need to validate variable input against a regular expression have no built-in way to do so. The only options were the fixed-format validators (`required`, `email`, `url`, `alpha`, `digit`, `alphanumeric`, `countrycode2`, `semver`, `length-{min}-{max}`), which can't express constraints like "must be a valid S3 bucket name" or "must match a project ID format."

## Motivation

I'm adding a feature to [Gruntwork Runbooks](https://github.com/gruntwork-io/runbooks) that will dynamically render a runbook based on an OpenTofu/Terraform module. As part of that functionality, it looks at the Tofu variable's `validation` block, and maps those to boilerplate validations. All conventional validations are already supported by boilerplate, except for the [regex](https://opentofu.org/docs/language/functions/regex/) function. Adding this to boilerplate allows for a clean mapping.

_Note: OpenTofu variable valudation works by specifying any boolean expression, so really any Tofu function is fair game, though in practice there are some heavily used patterns and `regex()` is understood to be one of them._

## Additional details

- It looks like we're writing more new code than we are. A large part of the change is carving out a `ConvertValidationStringtoRules()` helper function (so that `unmarshalValidationsField()` can call it directly).
- Boilerplate variable input will validate against a Go-compatible regular expression at prompt time, providing real-time feedback just like existing validators.
- If the user specifies a regex validation as a scalar string (versus a YAML list), then spaces in the regex aren't permitted because strings could still contain more than one pattern and are space-delimited. In the unlikely event the user does this, we return a helpful error message. 
- The regex pattern is compiled with `regexp.Compile` and matched via `validation.Match()` from the existing [go-ozzo/ozzo-validation](https://github.com/go-ozzo/ozzo-validation) dependency, so no new dependencies are introduced.
- These changes are fully backward-compatible. 

## TODOs

- [X] Update the docs.
- [X] Run the relevant tests successfully, including pre-commit checks.
- [X] Ensure any 3rd party code adheres with our [license policy](https://www.notion.so/gruntwork/Gruntwork-licenses-and-open-source-usage-policy-f7dece1f780341c7b69c1763f22b1378) or delete this line if its not applicable.
- [X] Include release notes. If this PR is backward incompatible, include a migration guide.

### `regex(pattern)` validation type

Added a new `regex(pattern)` validation type that lets users validate variables against arbitrary regular expression patterns at prompt time, with real-time feedback just like existing validators.

```yaml
variables:
  - name: BucketName
    description: Name of the S3 bucket
    type: string
    validations:
      - required
      - regex(^[a-z0-9][a-z0-9.-]*[a-z0-9]$)
```

`regex(pattern)` can be combined with any existing validator (`required`, `length-`, etc.) in the same `validations` list, and patterns are case-sensitive (e.g., `regex(^[A-Z]{2}-\d{4}$)` correctly distinguishes `AB-1234` from `ab-1234`).

## What's Changed

* feat: Add `regex(pattern)` validation type for matching variables against regular expressions

### DIFF
--- a/README.md
+++ b/README.md
@@ -748,7 +748,6 @@ variables:
  - name: ProfileName
    order: 2
    description: Enter the display name for your user
-
 ```
 
 #### Alternative Template Engines (EXPERIMENTAL)


variables/yaml_helpers.go
```diff
@@ -205,87 +207,142 @@ func (c CustomValidationRule) DescriptionText() string {
...
-func parseRuleString(ruleString string) []string {
-	ruleString = strings.ReplaceAll(ruleString, "]", "")
-	ruleString = strings.ReplaceAll(ruleString, "[", "")
-	ruleString = strings.ToLower(ruleString)
+// parseRuleString converts a space-delimited string of validation rules into a slice of
+// normalized rule strings. Used by ConvertValidationStringtoRules for its public API.
+// Because this splits on spaces, regex patterns containing spaces cannot be represented
+// in a space-delimited string. If a split regex pattern is detected, an error is returned
+// advising the user to use the YAML list format instead.
+func parseRuleString(ruleString string) ([]string, error) {
```
yhakbar: I think we should plan to just break space delimited support so that there's one consistent and reliable way to provide multiple validations.

variables/yaml_helpers.go
```diff
@@ -205,87 +207,142 @@ func (c CustomValidationRule) DescriptionText() string {
...
+
+	parts := strings.Split(ruleString, " ")
+	for i, part := range parts {
+		parts[i] = normalizeRuleString(part)
+	}
+
+	// Detect regex patterns that were broken by the space split: a token starting with
+	// "regex(" but not ending with ")" means the pattern contained a space.
+	for _, part := range parts {
+		if strings.HasPrefix(part, "regex(") && !strings.HasSuffix(part, ")") {
```
yhakbar: This part in specific makes me concerned. It feels brittle (e.g. if someone places a space after `regex` "`regex (...)`"

variables/yaml_helpers.go
```diff
@@ -205,87 +207,142 @@ func (c CustomValidationRule) DescriptionText() string {
...
-	return strings.Split(ruleString, " ")
+	return parts, nil
+}
+
+// convertSingleValidationRule converts a single validation rule string into a CustomValidationRule.
+// The rule string should already be normalized (lowercased for non-regex rules).
+func convertSingleValidationRule(rule string) (CustomValidationRule, error) {
+	switch {
+	case strings.HasPrefix(rule, "length-"):
+		valString := strings.TrimLeft(rule, "length-")
```
yhakbar: This is a bug. We want `strings.TrimPrefix` or `strings.CutPrefix`.

variables/yaml_helpers.go
```diff
@@ -205,87 +207,142 @@ func (c CustomValidationRule) DescriptionText() string {
...
 
-	return strings.Split(ruleString, " ")
+	return parts, nil
+}
+
+// convertSingleValidationRule converts a single validation rule string into a CustomValidationRule.
+// The rule string should already be normalized (lowercased for non-regex rules).
+func convertSingleValidationRule(rule string) (CustomValidationRule, error) {
+	switch {
+	case strings.HasPrefix(rule, "length-"):
```
yhakbar: I didn't know that our length validation rule had this syntax. 

I think if we're going to introduce a functional style validation for `regex()`, we should update the implementation of `length` so that it's also functional (e.g. `length(min, max)`). We can either update the implementation so that both are supported, or accept that this is a breaking change.

variables/yaml_helpers.go
```diff
@@ -205,87 +207,142 @@ func (c CustomValidationRule) DescriptionText() string {
...
+		return CustomValidationRule{
+			Validator: is.CountryCode2,
+			Message:   "Must be a valid ISO3166 Alpha 2 Country code",
+		}, nil
+	case rule == "semver":
+		return CustomValidationRule{
+			Validator: is.Semver,
+			Message:   "Must be a valid semantic version",
+		}, nil
+	case strings.HasPrefix(rule, "regex(") && strings.HasSuffix(rule, ")"):
```
yhakbar, line 287: NIT: We might want to trim whitespace before doing this, and have support for whitespace between the function name and the open paren.

README.md
```diff
@@ -728,6 +728,7 @@ Boilerplate uses the [`go-ozzo/ozzo-validation` library](https://github.com/go-o
  - "digit" - field must contain digits only
  - "countrycode2" - field must be an ISO3166 Alpha 2 Country code
  - "semver" - field must be a valid semantic version
+ - "regex(pattern)" - field must match the given regular expression pattern (e.g., `regex(^[a-z0-9-]+$)`)
```
yhakbar: I think we're better off with the pattern being an explicit string (e.g. `regex("^[a-z0-9-]+$")`).

If we have validation rules that accept parameters, we should make them explicitly typed.

variables/yaml_helpers.go
```diff
@@ -205,87 +207,142 @@ func (c CustomValidationRule) DescriptionText() string {
...
+		}, nil
+	default:
+		return CustomValidationRule{}, nil
+	}
+}
+
+// normalizeRuleString normalizes a single validation rule string by lowercasing it,
+// except for regex patterns which preserve their original case.
+// e.g. "Required" becomes "required" but "regex(^[A-Z]+$)" remains "regex(^[A-Z]+$)"
+func normalizeRuleString(rule string) string {
```
yhakbar, line 364: I think this is going to shoot us in the foot in the future. Why not just require proper casing for validations? We're eventually going to introduce a validation that accepts a string and forget to add an exception here.

variables/yaml_helpers.go
```diff
@@ -296,21 +353,59 @@ func ConvertValidationStringtoRules(ruleString string) ([]CustomValidationRule,
...
+//	  - regex(^[a-z ]+$)
 func unmarshalValidationsField(fields map[string]any) ([]CustomValidationRule, error) {
 	validations := fields["validations"]
+	if validations == nil {
+		return nil, nil
+	}
 
-	validationsAsString := fmt.Sprintf("%v", validations)
+	switch v := validations.(type) {
+	case []interface{}:
```
yhakbar: NIT: Prefer `any`

variables/yaml_helpers.go
```diff
@@ -296,21 +353,59 @@ func ConvertValidationStringtoRules(ruleString string) ([]CustomValidationRule,
...
+		// List of rules (e.g., ["required", "regex(^[a-z ]+$)"])
+		// Process each element individually to preserve spaces and brackets in patterns
+		var allRules []CustomValidationRule
```
yhakbar: NIT: This can be preallocated to len(v)

variables/yaml_helpers.go
```diff
@@ -296,21 +353,59 @@ func ConvertValidationStringtoRules(ruleString string) ([]CustomValidationRule,
...
 
-	rules, err := ConvertValidationStringtoRules(validationsAsString)
-	if err != nil {
-		return []CustomValidationRule{}, err
-	}
+		for _, item := range v {
+			rule := normalizeRuleString(fmt.Sprintf("%v", item))
```
yhakbar: NIT: We might want to assert that `item` is a string or something that implements the Stringer interface instead of doing this and logging a warning when it doesn't. I don't know that it makes a big deal either way.

variables/yaml_helpers.go
```diff
@@ -296,21 +353,59 @@ func ConvertValidationStringtoRules(ruleString string) ([]CustomValidationRule,
...
+
+			cvr, err := convertSingleValidationRule(rule)
+			if err != nil {
+				return nil, err
+			}
+
+			if cvr != (CustomValidationRule{}) {
```
yhakbar: NIT: Why are we doing this? We only hit this if we encounter the default return of `convertSingleValidationRule`, right? Does that mean that `CustomValidationRule{}` isn't a valid return value? Is `allRules` actually `allCustomRules`? I'm not sure this matters if it's working.

variables/yaml_helpers.go
```diff
@@ -296,21 +353,59 @@ func ConvertValidationStringtoRules(ruleString string) ([]CustomValidationRule,
...
+				allRules = append(allRules, cvr)
+			}
+		}
+
+		return allRules, nil
+	case string:
```
yhakbar, line 414: I feel like we should break this instead of continuing to support it. I'm not sure we gain a lot by having backwards compatibility for both having a single string as the validation and an array of validations.

variables/yaml_helpers.go
yhakbar: It's curious that we either support a string or a slice of any. It seems like we should either support any / slice of any or string / slice of string.

variables/yaml_helpers_test.go
```diff
@@ -212,13 +212,280 @@ func TestParserulestring(t *testing.T) {
...
+	t.Run("unrecognized scalar returns no rules", func(t *testing.T) {
+		t.Parallel()
+
+		fields := map[string]any{
+			"validations": "notarealrule",
+		}
+
+		rules, err := unmarshalValidationsField(fields)
+		require.NoError(t, err)
+		assert.Nil(t, rules)
```
yhakbar: NIT: Prefer `assert.Empty` for this kind of thing. Better reflects intent and will be resilient to any change that explicitly allocates an empty slice in the future instead of returning nil.

variables/yaml_helpers.go
josh-padnick: Fair enough! I've removed support and we now throw an error if the user tries to declare validations as a string.

variables/yaml_helpers.go
josh-padnick: I hear you. Ultimately, I went with strict parsing where you have to type `regex(...)` or `length(...)`, and `regex (...)` (with a space) will return an "unrecognized validation rule" error.

variables/yaml_helpers.go
josh-padnick: Good call. Fixed! Went with `strings.TrimPrefix` since I didn't need the return value.

variables/yaml_helpers.go
josh-padnick: Good call. I updated to support `length(min, max)`.

variables/yaml_helpers.go
josh-padnick: Fixed!

README.md
josh-padnick: Good call. I adopted this pattern.

variables/yaml_helpers.go
josh-padnick: Good call. I switched to requiring strict case matches.

variables/yaml_helpers.go
josh-padnick: Fixed!

variables/yaml_helpers.go
josh-padnick: Not a critical optimization but a fair point nonetheless, so adopted!

variables/yaml_helpers.go
josh-padnick: Good call. I switched to using types more strongly here.

variables/yaml_helpers.go
josh-padnick: Good call. This was a holdover from the previous function so I removed it.

variables/yaml_helpers.go
josh-padnick: Agreed. String rules now return an error.

variables/yaml_helpers.go
josh-padnick: The YAML parsing returns an `any[]` so there's no avoiding that.

variables/yaml_helpers_test.go
josh-padnick: Fixed!

josh-padnick:
@yhakbar Thanks for the excellent feedback. I adopted pretty much all your suggestions, with the one exception that I enforced no spaces between `regex` and `(`, so `regex()` is valid, but `regex ()` is not.

Take a final look at the changeset, and then I think we're all set here!

variables/yaml_helpers.go
```diff
@@ -205,112 +206,156 @@ func (c CustomValidationRule) DescriptionText() string {
...
-			max, maxErr := strconv.Atoi(strings.TrimSpace(maxStr))
-			if maxErr != nil {
-				return validationRules, maxErr
-			}
+		return CustomValidationRule{
+			Validator: validation.Length(min, max),
+			Message:   fmt.Sprintf("Must be between %d and %d characters long", min, max),
+		}, nil
+	case strings.HasPrefix(rule, `regex("`) && strings.HasSuffix(rule, `")`):
+		pattern := strings.TrimSuffix(strings.TrimPrefix(rule, `regex("`), `")`)
```
yhakbar: We need to think through escaping here. It's probable that someone will expect to be able to use something like the following as a regex pattern: `regex("They said: \"Hello world!\"")`.

We probably don't want to support `regex("They said: "Hello world!"")`, which I think would be supported in the implementation you have right now.

A common pattern that I think would work here is to support multiple string quoting characters and to support escaping quotes.

So someone could expect the following to work:

```
regex("They said: \"Hello world!\"")
regex(`They said: "Hello world!"`)
```

I think the simplest way to achieve this would be to use `strconv.Unquote` on the contents of `regex()` (without trimming the leading and closing quotes). That would make it so that any valid Golang string could be used as a value for `regex`.

variables/yaml_helpers.go
```diff
@@ -205,112 +206,156 @@ func (c CustomValidationRule) DescriptionText() string {
...
-				Message:   "Must be a valid ISO3166 Alpha 2 Country code",
-			}
-		case rule == "semver":
-			cvr = CustomValidationRule{
-				Validator: is.Semver,
-				Message:   "Must be a valid semantic version",
-			}
+		compiledRegex, err := regexp.Compile(pattern)
+		if err != nil {
+			return CustomValidationRule{}, fmt.Errorf("invalid regex pattern in validation 'regex(\"%s\")': %w", pattern, err)
```
yhakbar: If you accept the suggestion to use `strconv.Unquote`, make sure to update this error message to.

variables/yaml_helpers.go
```diff
@@ -205,112 +206,156 @@ func (c CustomValidationRule) DescriptionText() string {
...
+		}, nil
+	default:
+		return CustomValidationRule{}, fmt.Errorf("unrecognized validation rule %q", rule)
 	}
+}
 
-	return validationRules, nil
+// normalizeRuleString trims surrounding whitespace from a validation rule string.
+// Rule names and arguments are case-sensitive and must be written exactly as documented.
+func normalizeRuleString(rule string) string {
```
yhakbar, line 364: NIT: Is this really valuable? Could callers not just call `strings.TrimSpace` and have an easier time predicting behavior?

variables/yaml_helpers.go
```diff
@@ -205,112 +206,156 @@ func (c CustomValidationRule) DescriptionText() string {
...
+	if validations == nil {
+		return nil, nil
+	}
 
-	validationsAsString := fmt.Sprintf("%v", validations)
+	switch v := validations.(type) {
+	// We use []any (rather than []string) because Go's YAML libraries unmarshal
+	// sequences into []any, not []string, when the target type is any/interface{}.
+	case []any:
+		// List of rules (e.g., ["required", "regex(^[a-z ]+$)"])
```
yhakbar: NIT: This example isn't quoted in `regex()`

variables/yaml_helpers_test.go
```diff
@@ -188,37 +188,293 @@ func TestParserulestring(t *testing.T) {
...
+	return convertSingleValidationRule(normalizeRuleString(ruleString))
+}
+
+// TestConvertSingleValidationRule_Regex tests normalizeRuleString + convertSingleValidationRule
+// for regex rules. The production path goes through unmarshalValidationsField, but these tests
+// verify the string-parsing pipeline independently.
+func TestConvertSingleValidationRule_Regex(t *testing.T) {
+	t.Parallel()
+
+	t.Run("lowercase alphanumeric pattern passes", func(t *testing.T) {
```
yhakbar: NIT: This could be a test table.

variables/yaml_helpers_test.go
```diff
@@ -188,37 +188,293 @@ func TestParserulestring(t *testing.T) {
...
+		err = rule.Validator.Validate("ab-1234")
+		assert.Error(t, err)
+	})
+
+	t.Run("invalid regex returns error", func(t *testing.T) {
+		t.Parallel()
+
+		_, err := normalizeAndConvert(`regex("invalid[")`)
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "invalid regex pattern")
```
yhakbar: NIT: This is a smell. We should be using `assert.ErrorAs` or `assert.ErrorIs` here with a typed error.

yhakbar: Same with the other assertions against `err.Error()`.

README.md
yhakbar: This isn't accurate anymore. If you accept the suggestion to use `strconv.Unquote`, make sure to also call out that users can escape quotes and use back ticks here.

variables/yaml_helpers_test.go
josh-padnick: I hear you, but these are all user-facing config parsing errors produced by fmt.Errorf; we're not programmatically reacting to them right now. We could add typed errors, but that feels like a decent amount of work, so leaving this for a future improvement.

README.md
josh-padnick: Roger. Updated when I updated previous code.

variables/yaml_helpers.go
josh-padnick: Updated error message to show original rule as written, pre-parsing.

variables/yaml_helpers.go
josh-padnick: Yeah, I wrestled with this, too. Originally, it lower-cased rules, and then in the end all it did was trim whitespace. I figured I'd leave it in case we need to do other things to normalize in the future, but you could definitely argue it's over-engineered.

Leaving in for now since it's easy to change later and I'm eager to get this merged.

variables/yaml_helpers.go
josh-padnick: Good catch. Fixed.

variables/yaml_helpers_test.go
josh-padnick: Good catch. Updated.

variables/yaml_helpers.go
josh-padnick: That's a good point. I originally tried `strconv.Unquote`, but it interprets all Go escape sequences, which forces users to write `\\d` instead of `\d` in their regex patterns. This felt like leaking an implementation detail into the YAML syntax.

Instead, I AI-generated a custom `unquoteRegexPattern` function (with lots of unit tests to validate it) that only interprets `\"` and `\\` as escape sequences, passing everything else (like `\d`, `\w`, `\s`) through literally. I also added a check to reject unescaped interior quotes, so `regex("They said: "Hello world!"")` correctly produces an error, while `regex("They said: \"Hello world!\"")` works as expected. Both double-quoted and backtick-quoted patterns are supported.

PR #275, opened 2026-02-14
switch logging to slog

jlambert121:
Fixes https://github.com/gruntwork-io/boilerplate/issues/82.

Currently all logging is done via utils.Logger which just outputs logging to stdout. It would be nice to have more control over logging by silencing logging when not needed or injecting a separate logger (compatible with *slog.Logger).

There is a lot of lines of change here, but it's just creating a slog.Handler that prefixes [boilerplate] to log lines to maintain the existing look and passing an instance of the logger throughout. go is not my primary language though, any feedback is greatly appreciated!

Replaces #189

### DIFF
--- a/util/shell.go
+++ b/util/shell.go
@@ -1,6 +1,7 @@
 package util
 
 import (
+	"fmt"
 	"os"
 	"os/exec"
 	"strings"
@@ -10,7 +11,7 @@ import (
 
 // Run the given shell command with the given environment variables and arguments in the given working directory
 func RunShellCommandAndGetOutput(workingDir string, envVars []string, command string, args ...string) (string, error) {
-	Logger.Printf("Running command: %s %s", command, strings.Join(args, " "))
+	fmt.Printf("Running command: %s %s\n", command, strings.Join(args, " "))
 
 	cmd := exec.Command(command, args...)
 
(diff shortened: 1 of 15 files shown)


util/shell.go
```diff
@@ -1,57 +1,42 @@
...
-}
-
-// RunShellCommandAndGetOutputWithContext runs the given shell command with the given environment variables and arguments in the given working directory
-func RunShellCommandAndGetOutputWithContext(ctx context.Context, workingDir string, envVars []string, argslist ...string) (string, error) {
-	command := argslist[0]
-	args := argslist[1:]
-	Logger.Printf("Running command: %s %s", command, strings.Join(args, " "))
+// Run the given shell command with the given environment variables and arguments in the given working directory
+func RunShellCommandAndGetOutput(workingDir string, envVars []string, command string, args ...string) (string, error) {
+	fmt.Printf("Running command: %s %s\n", command, strings.Join(args, " "))
```
jlambert121, line 14: Not using `slog`, but seems like maybe it is OK for these?

jlambert121 (commented):
Sorry for the delay on this, rebased from `master` and incorporated feedback.  Let me know your thoughts!

Edit: Not sure why there are conflicts, investigating

PR #276, opened 2026-02-14
switch logging to slog

### LINKED ISSUE #82: Implement logger with logging levels

jlambert121:
Fixes https://github.com/gruntwork-io/boilerplate/issues/82.

Currently all logging is done via utils.Logger which just outputs logging to stdout. It would be nice to have more control over logging by silencing logging when not needed or injecting a separate logger (compatible with *slog.Logger).

There is a lot of lines of change here, but it's just creating a slog.Handler that prefixes [boilerplate] to log lines to maintain the existing look and passing an instance of the logger throughout. go is not my primary language though, any feedback is greatly appreciated!

Replaces https://github.com/gruntwork-io/boilerplate/pull/189 and https://github.com/gruntwork-io/boilerplate/pull/275

Sorry for the PR noise, I missed that the default branch name changed and rebased off of the wrong branch.


jlambert121:
Rebased

jlambert121:
Bump on this again, please let me know if there's anything you'd like changed - or if you'd rather keep the current implementation!

jlambert121:
Looks like #311 was implemented to achieve this.

PR #277, opened 2026-02-19
feat: Adding WASM build target

yhakbar:
## Description

I was able to play around and get a WASM build target working pretty quickly.

Works pretty well:

Screenshot 2026-02-19 at 15 39 43

While going through, there were two issues that stuck out and I addressed them while I was there.

The WASM binary being produced was 83MB. That was significantly larger than it needed to be. After digging into it a bit, I was able to work out that a lot of that bloat was from imports of things like `go-getter` and `go-commons` in the critical path used for rendering templates when using Boilerplate as a library.

Those things weren't critical to usage of Boilerplate as a library, so I moved out the integration with `go-getter` to a package that wouldn't get imported by any code used  by `RenderTemplateFromString`. That prevented it from getting compiled into the WASM binary. This should also make Boilerplate a much lighter import for consumers that don't depend on `go-getter` functionality.

While doing that, I also removed the dependency on `go-commons`, `go-errors`, and all of the transient dependencies they pulled in. At that stage, the binary size had dropped to 15MB.

After that, I also addressed `gopls` findings to modernize the codebase a bit, as we don't address that very often here.

Also cut out the color dependencies, `go-cmp`, and did some reorganization to make it easier to avoid importing bloat into the WASM binary. Size dropped to 13MB.

Cut a pre-release here:
https://github.com/gruntwork-io/boilerplate/releases/tag/v0.12.0-alpha1

With brotli compression it drops to 2MB.

https://github.com/gruntwork-io/boilerplate/releases/tag/v0.12.0-alpha3

## TODOs

- [x] Update the docs.

## Release Notes (draft)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,6 +45,9 @@ jobs:
           <<: *install_gruntwork_utils
       - run: build-go-binaries --app-name boilerplate --dest-path bin --ld-flags "-X github.com/gruntwork-io/boilerplate/version.Version=$CIRCLE_TAG"
       - run:
+          name: Install brotli
+          command: sudo apt-get update && sudo apt-get install -y brotli
+      - run:
           name: Build WASM binary
           command: |
             GOOS=js GOARCH=wasm go build -o bin/boilerplate.wasm -ldflags "-s -w -X github.com/gruntwork-io/boilerplate/version.Version=$CIRCLE_TAG" ./cmd/wasm/


cmd/wasm/main.go
```diff
@@ -0,0 +1,54 @@
...
+	var variables map[string]any
+	if err := json.Unmarshal([]byte(varsJSON), &variables); err != nil {
+		return js.Global().Get("Error").New(fmt.Sprintf("failed to parse variables JSON: %v", err))
+	}
+
+	opts := &options.BoilerplateOptions{
+		NonInteractive: true,
+		NoShell:        true,
+		OnMissingKey:   options.ExitWithError,
+		OnMissingConfig: options.Ignore,
```
denis256, line 45: Should we initialize ShellCommandAnswers field?

internal/fileutil/fileutil.go
```diff
@@ -0,0 +1,65 @@
...
+	return filepath.Walk(srcFolder, func(path string, info os.FileInfo, _ error) error {
+		relPath, err := filepath.Rel(srcFolder, path)
+		if err != nil {
+			return err
+		}
+
+		if IsDir(path) {
+			const defaultDirPerm = 0755
+			return os.MkdirAll(filepath.Join(targetFolder, relPath), defaultDirPerm)
+		} else {
```
denis256, line 53: Redundant 'else' in 'if'

internal/color/color.go
```diff
@@ -0,0 +1,16 @@
+// Package color provides lightweight ANSI terminal color helpers.
+package color
+
+const (
+	ansiReset   = "\033[0m"
```
denis256, line 5: Will this work on WASM environment?

yhakbar: It doesn't matter. It doesn't get imported or used there.

cmd/wasm/main.go
yhakbar: I don't think it can be used, but I can double-check.

PR #285, opened 2026-02-24
feat: Adding manifest

### LINKED ISSUE #224: Add list of generated files as optional output

yhakbar:
## Description

Fixes #224.

Closes #225.

Supports optional generation of a manifest file that tracks generated content from a template.

## TODOs

- [ ] Update the docs.

## Release Notes (draft)


internal/manifest/manifest.go
```diff
@@ -0,0 +1,148 @@
...
+	h := sha256.New()
+	if _, err := io.Copy(h, f); err != nil {
+		return "", err
+	}
+
+	return "sha256:" + hex.EncodeToString(h.Sum(nil)), nil
+}
+
+// GenerateSchema returns a JSON Schema for the Manifest type.
+func GenerateSchema() *jsonschema.Schema {
```
denis256, line 119: Looks like it is called only from tests now

internal/manifest/manifest.go
```diff
@@ -0,0 +1,148 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/invopop/jsonschema"
+	"gopkg.in/yaml.v2"
```
denis256: Make sense to use `yaml.v3` ?

internal/manifest/manifest.go
yhakbar: Ya, that's intentional. The CLI doesn't need to be able to generate it at the moment, I don't think.

docs/src/content/docs/advanced/manifest.mdx
```diff
@@ -0,0 +1,177 @@
...
+
+The manifest contains the following fields:
+
+| Field | Description |
+|-------|-------------|
+| `SchemaVersion` | URL pointing to the [Manifest Schema](#manifest-schema) for the manifest format |
+| `Timestamp` | UTC timestamp of the generation run (RFC 3339) |
+| `TemplateURL` | The `--template-url` value used for this run |
+| `BoilerplateVersion` | Version of boilerplate that produced the output |
+| `SourceChecksum` | Checksum of the template source. For git sources: `git-sha1:<commit>` or `git-sha256:<commit>`. For local sources: `sha256:<hex>` |
```
Resonance1584, line 64: Is it possible to add `SourceChecksum` to Dependencies as well?

docs/src/content/docs/advanced/manifest.mdx
```diff
@@ -0,0 +1,177 @@
...
+| `Variables` | User-defined template variables used during generation (builtin variables are excluded) |
+| `Dependencies` | Array of dependencies that were processed (or skipped) during the run |
+| `Dependencies[].Name` | Name of the dependency |
+| `Dependencies[].TemplateURL` | Resolved template URL for the dependency |
+| `Dependencies[].OutputFolder` | Resolved output folder for the dependency |
+| `Dependencies[].Skip` | Rendered skip expression (present even for skipped dependencies) |
+| `Dependencies[].ForEach` | List of `for_each` values used for this entry |
+| `Dependencies[].ForEachReference` | The `for_each_reference` variable name, if set |
+| `Dependencies[].VarFiles` | Variable files specified on the dependency |
+| `Dependencies[].Variables` | Resolved variable values passed to the dependency |
```
Resonance1584, line 76: Probably my ignorance here - but is this just the variables used by that dependency or is it all variables? I thought boilerplate passed a map of all variables to every dep.

docs/src/content/docs/advanced/manifest.mdx
```diff
@@ -0,0 +1,177 @@
...
+| `SourceChecksum` | Checksum of the template source. For git sources: `git-sha1:<commit>` or `git-sha256:<commit>`. For local sources: `sha256:<hex>` |
+| `OutputDir` | The `--output-folder` value used for this run |
+| `Variables` | User-defined template variables used during generation (builtin variables are excluded) |
+| `Dependencies` | Array of dependencies that were processed (or skipped) during the run |
+| `Dependencies[].Name` | Name of the dependency |
+| `Dependencies[].TemplateURL` | Resolved template URL for the dependency |
+| `Dependencies[].OutputFolder` | Resolved output folder for the dependency |
+| `Dependencies[].Skip` | Rendered skip expression (present even for skipped dependencies) |
+| `Dependencies[].ForEach` | List of `for_each` values used for this entry |
+| `Dependencies[].ForEachReference` | The `for_each_reference` variable name, if set |
```
Resonance1584, line 74: Q: Are these dynamic at runtime - or are they defined in the parent template?

I'm not sure I'm qualified to comment with my lack of knowledge of boilerplate - but if I can derive these values by following the parent TemplateURL and SourceChecksum then we can probably omit here?

internal/manifest/manifest.go
```diff
@@ -0,0 +1,147 @@
...
+// GenerateSchemaJSON returns the canonical JSON encoding of the manifest schema.
+// This is the authoritative output that the on-disk schema.json must match.
+func GenerateSchemaJSON() ([]byte, error) {
+	schema := GenerateSchema()
+
+	var buf bytes.Buffer
+
+	enc := json.NewEncoder(&buf)
+	enc.SetIndent("", "  ")
+	enc.SetEscapeHTML(false)
```
Resonance1584, line 140: Do we want escapehtml on the actual manifest encoding too on L85?

internal/fileutil/fileutil.go
```diff
@@ -41,7 +42,7 @@ func WriteFileWithSamePermissions(source string, destination string, contents []
 
 // CopyFolder copies all the files and folders in srcFolder to targetFolder.
 func CopyFolder(srcFolder string, targetFolder string) error {
-	return filepath.Walk(srcFolder, func(path string, info os.FileInfo, _ error) error {
+	return filepath.WalkDir(srcFolder, func(path string, _ fs.DirEntry, _ error) error {
```
Resonance1584, line 45: Not your fault - but I think we should be checking for the error arg in this function

docs/src/content/docs/advanced/manifest.mdx
yhakbar: Dependencies can have variables that are computed from the dependent template.

docs/src/content/docs/advanced/manifest.mdx
yhakbar: Boilerplate functions can have side effects, which means you can author templates with non deterministic values here. E.g. you use `shell` 
https://boilerplate.gruntwork.io/template-syntax/helper-functions/#file--shell-functions

Most of the time you really shouldn't, but you can.

internal/manifest/manifest.go
yhakbar: I'll remove it in a subsequent PR. I added it initially because of the schema contained gator brackets, but I since removed that text so that we don't need it anymore.

internal/fileutil/fileutil.go
yhakbar: I can address in a future PR.

PR #313, opened 2026-05-04
feat: input file mapping

odgrim:
## Description

Adds a new `boilerplate inputs map` subcommand (CLI) and `boilerplateInputsMap` WASM export that statically analyze a template tree and produce a JSON map of `declared input → output files affected`, plus an inverse `file → inputs` index. The analyzer parses templates into ASTs (no execution), expands partials transitively, walks dependencies, and propagates parent → child variable flow through both explicit value expressions and name-match inheritance.

This unblocks live-diff UIs and CI flows that need to know which files to re-render when a single input changes.

The PR also splits the interactive prompt and `ozzo-validation` code behind `!(js && wasm)` build tags so the WASM binary stops pulling terminal libraries it can't use.

## Highlights

- New `inputs` package with `FromOptions` (CLI, supports go-getter / disk) and `FromFS` (pure, WASM-safe) entry points sharing a single `analyze` core.
- Soft-error model: hard errors abort with non-zero exit; recoverable issues (`undeclared_variable`, `cycle`, `unresolvable_dependency`, `filename_render`, `parse`, `skip_files`, `partial_expansion_limit`) accumulate in `Result.Errors`.
- Honors `skip_files` (using `zglob` on disk to match runtime, `fs.Glob` in FS mode with a soft error for `**` patterns).
- Cycle detection on the dependency graph, with deterministic JSON output (sorted file lists, sorted error array).
- New docs page at `cli/inputs-map.mdx` describing the JSON schema and behavior.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Added a `boilerplate inputs map` subcommand and matching `boilerplateInputsMap` WASM export that statically analyze a template and report which output files are affected by each declared input variable. Refactored the WASM build to omit interactive-prompt and validation dependencies.

### Behavior notes

- **WASM input validation**: `validations` rules declared on variables are not executed in the WASM build (the validation package stub records the rules but no longer runs them, since `ozzo-validation` is now CLI-only). The CLI build is unchanged.
- **`inputs map` is read-only**: the subcommand never renders or writes files. It exits non-zero only on unrecoverable parse/download failures; recoverable issues land in the JSON `errors` array.

### Migration Guide

No breaking changes for CLI users. Library consumers building the WASM target previously got a no-op interactive prompt path through the `config` package — those callers should set `NonInteractive: true` (which the WASM bridge already does) and rely on `--var` / `--var-file` exclusively.


cli/boilerplate_cli.go
```diff
@@ -58,6 +58,7 @@ func CreateBoilerplateCli() *cli.App {
 	app.UsageText = "boilerplate [OPTIONS]"
 	app.Version = version.GetVersion()
 	app.Action = runApp
+	app.Commands = []*cli.Command{addInputsCommand()}
```
yhakbar: If we're gonna be adding new commands, we should add a `generate` command, and have it be the default command.

cli/inputs_cmd.go
```diff
@@ -0,0 +1,173 @@
...
+	"github.com/gruntwork-io/boilerplate/inputs"
+	"github.com/gruntwork-io/boilerplate/options"
+	"github.com/gruntwork-io/boilerplate/pkg/logging"
+	"github.com/gruntwork-io/boilerplate/variables"
+)
+
+const inputsMapHelpText = `Usage: boilerplate inputs map [OPTIONS]
+
+Analyze the template at --template-url and print a JSON object describing which
+output files are affected by each declared input variable. The command never
```
yhakbar, line 22: What if the variable is used for determining the filepath?

cli/inputs_cmd.go
```diff
@@ -0,0 +1,173 @@
...
+
+The command exits with a non-zero status only on unrecoverable parse failures.
+Soft errors (e.g., a referenced variable that is not declared in any
+boilerplate.yml in scope) appear in the "errors" array and do not change the
+exit code.`
+
+// addInputsCommand returns the "inputs" command with its "map" subcommand.
+// It is exported only via CreateBoilerplateCli; defining it here keeps the
+// subcommand wiring out of the main CLI file.
+func addInputsCommand() *cli.Command {
```
yhakbar: This is actually a constructor for the inputs command, right? It should be `newInputsCommand` if so.

cli/inputs_cmd.go
odgrim: it does get a bit wonky here where not supplying defaults starts giving empty partial-paths like:
    "management/_global/-role/terragrunt.hcl": [
      ".:ApplyIAMRoleName",
      ".:PlanIAMRoleName",
      "management/_global/-role:ApplyIAMRoleName",
      "management/_global/-role:GitLabOidcRoleModuleUrl",
      "management/_global/-role:GitLabOidcRoleModuleVersion",
      "management/_global/-role:RootTerragruntFileName"
    ],

that said, that is what it would resolve to when empty so it feels correct.

cli/boilerplate_cli.go
odgrim: I propose for reviewability we punt on this.

Reworking boilerplate and giving it the cloudnuke makeover treatment feels correct but beyond scope of this change. Thoughts?

yhakbar: IDK what this has to do with CloudNuke, but I'm not trying to block your progress.

yhakbar (dismissed):
LGTM

cmd/wasm/main.go
```diff
@@ -53,3 +57,71 @@ func renderTemplate(this js.Value, args []js.Value) any {
...
+//	{ "rootPath": ".", "files": { "<path>": "<contents>", ... } }
+//
+// The bundle must include every boilerplate.yml in the dependency tree,
+// keyed by its path relative to rootPath. Remote dependency template-urls
+// are not resolvable in WASM and produce "unresolvable_dependency" entries
+// in the result's errors array.
+func inputsMap(this js.Value, args []js.Value) any {
+	defer func() {
+		if r := recover(); r != nil {
+			fmt.Println("boilerplate: recovered from panic in inputsMap:", r)
```
yhakbar: This should write to stderr

cmd/wasm/main.go
```diff
@@ -53,3 +57,71 @@ func renderTemplate(this js.Value, args []js.Value) any {
...
+	}()
+
+	if len(args) < 2 {
+		return js.Global().Get("Error").New("boilerplateInputsMap requires 2 arguments: bundleJSON, varsJSON")
+	}
+
+	bundleJSON := args[0].String()
+	varsJSON := args[1].String()
+
+	var bundle struct {
```
yhakbar: This should be an external struct, not something inline here.

cmd/wasm/main.go
```diff
@@ -53,3 +57,71 @@ func renderTemplate(this js.Value, args []js.Value) any {
...
+	for p, contents := range bundle.Files {
+		mfs[p] = &fstest.MapFile{Data: []byte(contents)}
+	}
+
+	var variables map[string]any
+	if err := json.Unmarshal([]byte(varsJSON), &variables); err != nil {
+		return js.Global().Get("Error").New(fmt.Sprintf("failed to parse variables JSON: %v", err))
+	}
+
+	result, err := inputs.FromFS(context.Background(), mfs, rootPath, variables)
```
yhakbar: This context should be defined earlier and passed in.

config/get_variables_prompt.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+
+		prompt := &survey.Input{
+			Message: msg,
+		}
+
+		err := survey.AskOne(prompt, &value)
+		if err != nil {
+			if errors.Is(err, terminal.InterruptErr) {
+				log.Fatal("quit")
```
yhakbar: Avoid using `log.Fatal`

yhakbar (dismissed):

